Issue #5, opened 2015-09-17
support an app update user flow


mykmelez:
Oghliner should support an app update user flow, so an app that uses Oghliner receives a message when an update is available and can prompt the user to update the app (if it chooses to do so, and at the appropriate time).

The flow would be optional, so an app that wants updates to be completely automatic and silent can simply ignore the message and never prompt the user to update the app, which will then update itself automatically whenever the service worker is updated by the browser (which is currently the default and only supported flow).

See [sw-precache's demo service-worker-registration.js file](https://github.com/GoogleChrome/sw-precache/blob/7688ee8ccdaddd9171af352384d04d16d712f9d3/demo/app/js/service-worker-registration.js) for inspiration, but ideally we would not have apps add their custom logic to our own equivalent (service-manager.js) but rather expose the registration to their code in a way that makes it possible for them to listen for its events and handle them in their own script.

mykmelez:
@delapuente I wonder if this is something that will be easier to do with _offliner_ than with _sw-precache_.

delapuente:
According to the _offliner_ model, when an update is available (the case for oghliner, as soon as the worker changes), an evolve strategy executes (this case a reinstall is taking effect) and an `activationPending` is issued to the client. The activation swap the offline caches.

The default behaviour for oghliner mimics what is implemented now so, we listen to `activationPending` and reload but we could default into simply ignore the signal and show a tip for implementation and add the reload only if the developer use the option `--on-update reload` or something like this. The opposite is interesting as well, autoreload by default providing a complete sample lifecycle and use some like `--on-update ignore` to manually handle the activation.

marco-c:
So by using offliner (#96) we basically get this for free (because the user can define his update strategy), don't we?

delapuente:
Yep, you have the current implementation here: https://github.com/mozilla/oghliner/pull/96/files#diff-1e6990c753cb6c767172ea5f54207c2fR6

mykmelez:
> So by using offliner (#96) we basically get this for free (because the user can define his update strategy), don't we?

If so, then it may be that the only thing to do here, once #96 lands, is to document the support for an app update flow. But I want to make sure that the flow we get with #96 is the one described by this issue, in particular (from the original description):

> Oghliner should support an app update user flow, so an app that uses Oghliner receives a message when an update is available and can prompt the user to update the app (if it chooses to do so, and at the appropriate time).

Also:

[...]

PR #11, opened 2015-09-23
Add prompt when an updated service worker is found.

brendandahl:
Simple version before we start adding a "pretty" UI.

### DIFF
--- a/app/js/offline-manager.js
+++ b/app/js/offline-manager.js
@@ -1,5 +1,29 @@
+function updateFound() {
+  var installingWorker = this.installing;
+
+  // Wait for the new service worker to be installed before prompting to update.
+  installingWorker.addEventListener('statechange', function() {
+    switch (installingWorker.state) {
+      case 'installed':
+        // Only show the prompt if there is currently a controller so it is not
+        // shown on first load.
+        if (navigator.serviceWorker.controller &&
+            window.confirm('An updated version of this page is available, would you like to update?')) {
+          window.location.reload();
+          return;
+        }
+        break;
+
+      case 'redundant':
+        console.error('The installing service worker became redundant.');
+        break;
+    }
+  });
+}
+
 if ('serviceWorker' in navigator) {
   navigator.serviceWorker.register('offline-worker.js').then(function(registration) {
     console.log('offline worker registered');
+    registration.addEventListener('updatefound', updateFound);
   });
 }


marco-c:
Instead of actually doing something, we could call a user-defined function to let them do what they want (I guess, depending on the use case, a developer could want to force the update or ignore it, instead of prompting the user).

mykmelez:
Argh, GitHub didn't make me watch this repository after I moved it from my personal account to the @mozilla organization, so I only just now saw this pull request.

> Instead of actually doing something, we could call a user-defined function to let them do what they want (I guess, depending on the use case, a developer could want to force the update or ignore it, instead of prompting the user).

I like that idea. Then the Oghliner app could implement that user-defined function as a demo of how to do it, and a developer could reuse Oghliner's implementation or create their own.

But this is a good start, so I'm going to merge this, and then we can make that change in a followup.

marco-c:
> But this is a good start, so I'm going to merge this, and then we can make that change in a followup.

I was going to file a follow-up issue, but I think https://github.com/mozilla/oghliner/issues/5 serves the purpose.

brendandahl:
> Instead of actually doing something, we could call a user-defined function to let them do what they want (I guess, depending on the use case, a developer could want to force the update or ignore it, instead of prompting the user)

I had considered this, but I'm not sure of the value of it as it seems we'd just be re-implementing these events and the user could just listen to the events themselves and implement whatever they want. I think it would be good to have better UI for this when we make things "pretty."

marco-c:
> I had considered this, but I'm not sure of the value of it as it seems we'd just be re-implementing these events and the user could just listen to the events themselves and implement whatever they want.

That's true, the only value would be to hide the service workers API from users and provide a simpler API.

> I think it would be good to have better UI for this when we make things "pretty."

I'm not sure we should implement any UI in oghliner, in particular when using it as a tool. I'd expect it to simply allow me, as an app developer, to simplify adding offline support and deploying to GitHub pages, but I wouldn't want it to provide a different UI than what I want for my app.

mykmelez:
> I'm not sure we should implement any UI in oghliner, in particular when using it as a tool. I'd expect it to simply allow me, as an app developer, to simplify adding offline support and deploying to GitHub pages, but I wouldn't want it to provide a different UI than what I want for my app.

Agreed, when using it as a tool: then it should simplify offlining and deploying to GitHub Pages without modifying your app's look and feel. But when using it as a template, it's useful for it to provide an update flow that you can use as-is, repurpose, or remove.

PR #14, opened 2015-09-25
Allow importing custom scripts in the service worker script

marco-c:


mykmelez:
It'd be helpful to link "Push API" to https://developer.mozilla.org/en-US/docs/Web/API/Push_API, so users reading the docs can get more info.

mykmelez:
I added that link and merged this!

PR #17, opened 2015-09-30
Add boostrap command.

brendandahl:
Starting point for the boostrap command which creates the skeleton template app in any folder. The process for starting from template would now be

``` (bash)
git clone git@github.com:brendandahl/testogh.git
cd testogh
oghliner bootstrap
gulp build
gulp deploy
```

Or we could just do`oghliner deploy app` instead of the last two steps.

More things we could do:
- auto fill out package.json with data from git config (we should use node module gulp-template)
- prompt user for app name or maybe parse this from git url
- run bootstrap from oghliner to build dist directory


mykmelez:
After `oghliner bootstrap`, you also need to `npm install gulp` in order to do `gulp build`, even though you have global _gulp_, because _gulp_ evaluates _gulpfile.js_, which requires a local instance of _gulp_.

(If there was a package.json, this could be `npm install`.)

mykmelez:
(Then you need to `npm install gulp-connect` and `npm install oghliner`.)

mykmelez:
One other thought: if we copy the files to template/app/ rather than moving them there, then we can differentiate between the content of the template and the content of Oghliner's own website. It would also reduce the differences between the template's gulpfile.js and Oghliner's own version of that file, since Oghliner's own version could build from app/ instead of template/app/.

mykmelez:
Note: I made a few updates to the content of app/index.html and updated the file globs in gulpfile.js to offline all the files in app/ (it was missing some of them because I hadn't updated the globs after I used the GitHub Pages Generator to design a nicer layout for the site).

brendandahl:
> (Then you need to npm install gulp-connect and npm install oghliner.)

Missed my package file. The local dependencies should be installed by gulp-install.

So now I think we're at:

```
npm install -g gulp oghliner
git clone git@github.com:brendandahl/testogh.git
cd testogh
oghliner bootstrap
gulp offline
gulp deploy
```

gulpfile.js
```diff
@@ -23,7 +23,7 @@ var oghliner = require('./index.js');
 gulp.task('default', ['build', 'offline']);
 
 gulp.task('build', function(callback) {
-  return gulp.src('app/**').pipe(gulp.dest('dist'));
+  return gulp.src('templates/app/**').pipe(gulp.dest('dist'));
```
mykmelez: Now that you're no longer removing the app/ subdirectory, this should continue to build app/ rather than templates/app/.

mykmelez:
The output with gulp-conflict is a bit scary, since it puts the word `conflict` in every line:

```
10-06 17:28 > oghliner bootstrap
[17:29:05] [conflict] Creating app
[17:29:05] [conflict] Creating gulpfile.js
[17:29:05] [conflict] Creating package.json
[17:29:05] [conflict] Creating app/images
[17:29:05] [conflict] Creating app/index.html
[17:29:05] [conflict] Creating app/scripts
[17:29:05] [conflict] Creating app/styles
[17:29:05] [conflict] Creating app/images/bg_hr.png
[17:29:05] [conflict] Creating app/images/blacktocat.png
[17:29:05] [conflict] Creating app/images/icon_download.png
[17:29:05] [conflict] Creating app/images/sprite_download.png
[17:29:05] [conflict] Creating app/scripts/offline-manager.js
[17:29:05] [conflict] Creating app/styles/github-light.css
[17:29:05] [conflict] Creating app/styles/stylesheet.css
```

But it seems like a good idea to use it anyway.

mykmelez:
The other output I noticed are these warnings:

> npm WARN package.json oghliner-template-app@0.1.0 No description
> npm WARN package.json oghliner-template-app@0.1.0 No repository field.
> npm WARN package.json oghliner-template-app@0.1.0 No README data
> npm WARN package.json oghliner-template-app@0.1.0 No license field.

I'm not sure how to pick a license for the user, although perhaps NPM would be satisfied if we added the field and gave it an empty string value. That reminds me that we should probably put the template files into the public domain (except for the ones already copyrighted by GitHub, of course). I'll ping licensing to confirm that.

As for description and repository, we should be able to pick sensible values for those. And creating a blank README.md file probably satisfies the README warning.

brendandahl:
> The output with gulp-conflict is a bit scary,

My thought too, I'll look into removing it.

As for package.json, we could prompt the user during bootstrap for values.

mykmelez:
> npm install -g gulp oghliner
> git clone git@github.com:brendandahl/testogh.git
> cd testogh
> oghliner bootstrap
> gulp offline
> gulp deploy

The other thing to do is commit the newly created files to the master branch, something like:

> git add app/ gulpfile.js package.json && git commit -m"bootstrap app using Oghliner"

Which reminds me of two more things:
- We should include a template .gitignore to ignore the node_modules directory.
- We should update the README to document this new way to bootstrap an app (and stop recommending that users clone or merge the repo).

mykmelez:
> As for package.json, we could prompt the user during bootstrap for values.

Ah, good idea!

brendandahl:
- [ ] add .gitignore
- [ ] public domain files

lib/bootstrap.js
```diff
@@ -0,0 +1,132 @@
...
+    out.push(key + ": " + config[key]);
+  }
+  return out.join("\n");
+}
+
+function getDefaultTemplateConfig(dir) {
+  var config = {
+    name: "oghliner-template-app",
+    repository: "https://oghliner-template-app.git",
+    description: "A template app boostrapped with oghliner.",
```
mykmelez: Nit: boostrapped -> bootstrapped.

mykmelez:
Looks good! I heard back from licensing and filed #38 to release the template files to the public domain, so we can follow up on that there.

PR #18, opened 2015-09-30
Warn when the rootDir contains the node_modules directory

marco-c:
While working on https://github.com/marco-c/tictactoe/, I was using the default value for the `rootDir` option.
This was causing the entire `node_modules` directory to be uploaded to `gh-pages`, but most of the files in `node_modules` weren't actually needed (at the time, actually, no one was needed).

So, I think we should warn the user when oghliner is going to upload the entire `node_modules` directory.

In the future, we should probably allow the user to specify a file whitelist/blocklist.

### DIFF
--- a/lib/deploy.js
+++ b/lib/deploy.js
@@ -23,10 +23,17 @@ var ghPages = require('gh-pages');
 var gitconfiglocal = require('gitconfiglocal');
 var gutil = require('gulp-util');
 
+var fs = require('fs');
+var path = require('path');
+
 module.exports = promisify(function(config, callback) {
   config = config || {};
 
   var rootDir = config.rootDir ? config.rootDir : '.';
+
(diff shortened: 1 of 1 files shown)


mykmelez:
> In the future, we should probably allow the user to specify a file whitelist/blocklist.

Yes, indeed. Unfortunately, the _gh-pages_ module doesn't make this easy, as its _src_ config option only accepts a single [minimatch](https://github.com/isaacs/minimatch) string. There are ways to specify globs to exclude with minimatch, but I don't think it's possible to specify both includes and excludes in a single string.

In any case, I wouldn't want to make users jump through hoops to figure out how to craft a minimatch string. Better to do something similar to the _fileGlobs_ config option for the _offline_ task, which takes an array rather than a single string.

PR #41, opened 2015-10-09
Add mocha for tests.

brendandahl:
I started writing some tests for bootstrap, but thought I'd put this up if more people are adding tests or there's some objection to using mocha.


mykmelez:
Mocha seems like a good option to me.

@marco-c You mentioned that you wrote some testing code for your Push library that we might be able to reuse. How would that work with Mocha?

marco-c:
It should be very easy to port my tests to use Mocha!

mykmelez:
Woot, then let's merge this and use Mocha for testing.

PR #50, opened 2015-10-09
Set cacheID to GitHub slug or name from package.json

marco-c:


mykmelez:
Looks good. Note that you can use _promisify_ to wrap functions like _ghslug_ in a promise-returning wrapper. In this case, it doesn't really matter, because we only call two asynchronous functions. But we should consider that if we add another asynchronous call here.

PR #51, opened 2015-10-09
Use github-slug in configure

marco-c:
I haven't tested it yet.


marco-c:
There was a minor bug, now fixed!

mykmelez:
Nice. Works for me too locally.

PR #63, opened 2015-10-13
Add basic test for bootstrap.

### LINKED ISSUE #36: figure out how to test bootstrap command
We should figure out how to test the _bootstrap_ command. Like _offline_ (#22), its effects are all local, so this should be straightforward.

brendandahl:
I had some issues with the gulp-install not triggering a finish event, but that can be fixed with calling sink() at the end. However, I have install disabled for tests as it is slow and require network connectivity.

Fixes #36

### DIFF
--- a/lib/bootstrap.js
+++ b/lib/bootstrap.js
@@ -23,6 +23,8 @@ var install = require('gulp-install');
 var rename = require("gulp-rename");
 var inquirer = require('inquirer');
 var gitconfiglocal = require('gitconfiglocal');
+var through2 = require('through2');
+var util = require('gulp-util');
 
 function prompt(questions) {
   // inquirer doesn't follow the node style callback so we can't use promisify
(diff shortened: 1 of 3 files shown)


marco-c:
Looks good to me, I have a single question. Why do you need the _noInstall_ option? Does the test fail on Travis if you try to install the dependencies? Or did you add it because otherwise the test takes too much time?

lib/bootstrap.js
```diff
@@ -131,7 +154,8 @@ module.exports = function(config) {
           .pipe(template(templateConfig))
           .pipe(conflict(rootDir))
           .pipe(gulp.dest(rootDir))
-          .pipe(install());
+          .pipe(config.noInstall ? util.noop() : install())
```
marco-c: Nit: couldn't you directly use sink here instead of util.noop and then sink?

brendandahl: Install also doesn't work if you don't do a sink().

brendandahl:
> Why do you need the noInstall option?

Mainly because it's slow as mentioned above. Let me try on travis and see if how it's doing.

lib/bootstrap.js
marco-c: Ah, ok. Could you update the comment in the following line then?

marco-c:
> Mainly because it's slow as mentioned above

Oops, I've skimmed over the first comment! I really didn't notice you wrote something in it :smile:

marco-c:
There are a few _npm ERR! tar.unpack untar error_ and then the test times out.

PR #64, opened 2015-10-13
Avoid a failing test affecting other tests' results

marco-c:


marco-c:
Close in favor of #73, which is a better alternative.

PR #65, opened 2015-10-14
Allow running a single test with 'gulp test'

marco-c:
Example usage: `gulp test --file test/testDeploy.js`


mykmelez:
Can we simplify to `gulp test test/testDeploy.js`? That'll also generalize to running multiple (but a subset) of tests in the future with `gulp test test/testFoo.js test/testBar.js`.

marco-c:
If you run _gulp test test/testDeploy.js_, gulp thinks test/testDeploy.js is a task (it prints this error message: `Task 'test/testDeploy.js' is not in your gulpfile`).

The recipe in the gulp repo (https://github.com/gulpjs/gulp/blob/master/docs/recipes/pass-arguments-from-cli.md) also uses params passed with hypens, which makes me think it isn't possible/easy to do it otherwise.

mykmelez:
Ah, ok. That makes me wonder if we're better off telling people to `mocha test/testDeploy.js`. But we can still take this.

PR #68, opened 2015-10-14
Don't use fs.exists/fs.existsSync

### LINKED ISSUE #66: Replace fs.exists/fs.existsSync with fs.access/fs.accessSync or fs.stat/fs.statSync
fs.exists and fs.existsSync are deprecated: https://nodejs.org/api/fs.html#fs_fs_exists_path_callback

marco-c:
Fixes #66.


brendandahl:
What about using a library or writing our own function? try/catch isn't very readable.

marco-c:
We can use _fs.stat_ when try/catch makes things less readable.

marco-c:
This should be ready now, I've removed the _try/catch_ and used the async fs.stat instead (which also allows us to have more information about the file). I've left _try/catch_ in one place because I think it was needed anyway and I added a comment to explain why.

brendandahl:
needs a rebase, then it's good to go

mykmelez:
@brendandahl Ready to merge now that it's rebased?

brendandahl:
Yes, sorry ended up being out sick yesterday.

PR #69, opened 2015-10-14
Use the 'temp' module to create temporary directories in testIntegrate

marco-c:


marco-c:
I'm going to merge this since it's a really simple change and I'm working on another patch that is going to modify the same file.

PR #72, opened 2015-10-14
Test 'offline'

### LINKED ISSUE #22: figure out how to test the *offline* command
We need to figure out how to test all the commands. _deploy_ and _configure_ might be harder, because they interact with servers; but _offline_ at least should be doable, so let's start there.

marco-c:
Fixes #22.


test/testOffline.js
```diff
@@ -0,0 +1,149 @@
...
+describe('Offline', function() {
+  var oldWd = process.cwd();
+
+  afterEach(function() {
+    process.chdir(oldWd);
+    offline.__set__('ghslug', ghslug);
+  });
+
+  it('should create offline-worker.js in the destination directory', function() {
+    var dir = temp.mkdirSync('tmp');
```
mykmelez: Nit: a string argument to _mkdirSync_ is a prefix that is prepended to the name of the directory, and it's intended to distinguish that directory from others inside the system-wide temp directory, so a string like _oghliner_ would work better here (like this [example of a "PDF Creator"](https://github.com/bruce/node-temp/blob/master/examples/pdfcreator.js), which uses the prefix _pdfcreator_).

test/testOffline.js
```diff
@@ -0,0 +1,149 @@
...
+
+  it('should use importScript in the service worker if the importScripts option is defined', function() {
+    var dir = temp.mkdirSync('tmp');
+
+    return offline({
+      rootDir: dir,
+      importScripts: [ 'a-script.js', ],
+    }).then(function() {
+      var content = fs.readFileSync(path.join(dir, 'offline-worker.js'), 'utf8');
+      assert.notEqual(content.indexOf('importScripts("a-script.js");'), -1);
```
mykmelez, line 48: This is fine, but in the future, let's make it more robust by parsing the JS with something like [acorn](https://www.npmjs.com/package/acorn) and asserting that the expected call exists within the AST.

marco-c, line 48: I thought about this when I started writing the test, I wanted to use https://www.npmjs.com/package/esprima. Then I thought the test was very simple and it wasn't worth it, at least for this first iteration.

Issue #20, opened 2015-10-05
await sync if it's already in progress when configure tells Travis to sync


mykmelez:
When the _configure_ command doesn't find your repository in Travis, it tells Travis to sync it and then waits for the sync to complete (periodically polling Travis to find out the status of the sync). But if Travis is _already_ syncing your repository when _configure_ tells it to sync, then the _travis.users.sync.post_ call throws `{ message: 'Sync already in progress. Try again later.' }`, and the command doesn't catch that error, so it simply reports the error and dies:

> Checking the status of your repository in Travis…
> I didn't find your repository in Travis.  Syncing Travis with GitHub…
> { message: 'Sync already in progress. Try again later.' }

It should instead catch that error and call the _travisAwaitSyncing_ polling function in that case, just as it does in the case of the repository not being found.

marco-c:
I have a patch for this, but I will wait #60 to be merged.

PR #75, opened 2015-10-15
Await Travis sync if it's already in progress

marco-c:
This is still untested, I'm unsure how to reliably reproduce https://github.com/mozilla/oghliner/issues/20.

### DIFF
--- a/lib/configure.js
+++ b/lib/configure.js
@@ -310,10 +310,17 @@ module.exports = function() {
       });
     }
 
-    function travisAwaitSyncing() {
+    function travisIsSyncing() {
       return travis.users.get()
       .then(function(res) {
-        if (res.user.is_syncing) {
+        return res.user.is_syncing;
+      });
+    }
+
+    function travisAwaitSyncing() {
+      return travisIsSyncing()
+      .then(function(isSyncing) {
+        if (isSyncing) {
           process.stdout.write('Waiting for Travis to finish syncing…\n');
           return new Promise(function(resolve, reject) { setTimeout(resolve, 5000) })
           .then(travisAwaitSyncing);
@@ -326,6 +333,22 @@ module.exports = function() {
       if (err.message === 'repository not found') {
         process.stdout.write('I didn\'t find your repository in Travis.  Syncing Travis with GitHub…\n');
         return travis.users.sync.post()
+        .catch(function(err) {
+          // Ignore the exception if Travis is already syncing. The status code
+          // of the error is 409 and its message is 'Sync already in progress.
+          // Try again later.', but instead of relying on that we simply perform
+          // an explicit check.
+          return travisIsSyncing().then(function(is_syncing) {
+            if (!is_syncing) {
+              // Check again if it's active on Travis (in case it finished syncing
+              // right before we checked)
+              return ensureActiveInTravis().catch(function() {
+                // Throw the original error.
+                throw err;
+              });
+            }
+          });
+        })
         .then(travisAwaitSyncing)
         .then(ensureActiveInTravis);
       }
--- a/lib/configure.js
+++ b/lib/configure.js
@@ -346,8 +346,8 @@ module.exports = function() {
           // of the error is 409 and its message is 'Sync already in progress.
           // Try again later.', but instead of relying on that we simply perform
           // an explicit check.
-          return travisIsSyncing().then(function(is_syncing) {
-            if (!is_syncing) {
+          return travisIsSyncing().then(function(isSyncing) {
+            if (!isSyncing) {
               // Check again if it's active on Travis (in case it finished syncing
               // right before we checked)
               return ensureActiveInTravis().catch(function() {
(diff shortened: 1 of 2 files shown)


mykmelez:
> This is still untested, I'm unsure how to reliably reproduce #20.

I'm not sure how reliable it is, but if you load your [Travis profile page](https://travis-ci.org/profile) and then press the Sync button on that page before invoking `oghliner configure` (or at least before completing the authentication part of the flow), then you may be able to catch Travis in the process of syncing. Of course you also have to be configuring a repository that Travis doesn't already know about, which is usually one that you've just created.

mykmelez:
While doing all that, if you do it on the test-configure branch in #60 and uncomment the nock recording code at the top of configure.js, then you'll generate nock recording output that you can use to add a test for this into testConfigure.js.

marco-c:
I was able to reproduce it by adding multiple calls to _travis.users.sync.post_ in sequence.

marco-c:
I also collected two nock recordings, with and without this patch, so I can easily write a test for it.

marco-c:
I've added four tests (two success and two failure cases):
- travis.users.sync.post throws because Travis is already syncing, we check if Travis is syncing and it is.
- travis.users.sync.post throws because Travis is already syncing, we check if Travis is syncing and it isn't, so we check if the repository is active and it is.
- travis.users.sync.post throws because Travis is already syncing, we check if Travis is syncing and it isn't, so we check if the repository is active and it isn't.
- travis.users.sync.post throws because of a generic error, we check if Travis is syncing and it isn't, so we check if the repository is active and it isn't.

lib/configure.js
```diff
@@ -334,6 +341,22 @@ module.exports = function() {
       if (err.message === 'repository not found') {
         process.stdout.write('I didn\'t find your repository in Travis.  Syncing Travis with GitHub…\n');
         return travis.users.sync.post()
+        .catch(function(err) {
+          // Ignore the exception if Travis is already syncing. The status code
+          // of the error is 409 and its message is 'Sync already in progress.
+          // Try again later.', but instead of relying on that we simply perform
+          // an explicit check.
+          return travisIsSyncing().then(function(is_syncing) {
```
mykmelez: Nit: this function takes an _is_syncing_ parameter, but the equivalent one above takes _isSyncing_. They should use the same format.

lib/configure.js
```diff
@@ -334,6 +341,22 @@ module.exports = function() {
...
+              // Check again if it's active on Travis (in case it finished syncing
+              // right before we checked)
+              return ensureActiveInTravis().catch(function() {
+                // Throw the original error.
+                throw err;
+              });
+            }
+          });
+        })
         .then(travisAwaitSyncing)
```
mykmelez, line 360: Both the _catch_ function and _travisAwaitSyncing_ call _travisIsSyncing_, which means that _travisIsSyncing_ gets called twice in a row when Travis is already syncing, first by the _catch_ function and then by _travisAwaitSyncing_. That isn't the end of the world, but it does result in an extra, unnecessary network request.

Another issue is that _ensureActiveInTravis_ will get called twice if Travis finishes syncing right before the _catch_ function calls _travisIsSyncing_: first by the _catch_ function, and then by the _ensureActiveInTravis_ link at the end of the chain.

To resolve these problems, I would make _travisAwaitSyncing_ wait unconditionally and then recurse if Travis is still syncing:

``` js
    function travisAwaitSyncing() {
      process.stdout.write('Waiting for Travis to finish syncing…\n');
      return new Promise(function(resolve, reject) { setTimeout(resolve, 5000) })
      .then(travisIsSyncing)
      .then(function(isSyncing) {
        if (isSyncing) {
          return travisAwaitSyncing();
        }
      });
    }
```

Then I would make the _catch_ function explicitly ignore the `Sync already in progress…` error, so the promise chain resolution falls through to the existing _travisAwaitSyncing_ call in the chain:

``` js
    return ensureActiveInTravis()
    .catch(function(err) {
      if (err.message === 'repository not found') {
        process.stdout.write('I didn\'t find your repository in Travis.  Syncing Travis with GitHub…\n');
        return travis.users.sync.post()
        .catch(function(err) {
          if (/* status code is 409 && message is 'Sync already in progress… */) {
            // Ignore the exception if Travis is already syncing.
          } else {
            throw err;
          }
        })
        .then(travisAwaitSyncing)
        .then(ensureActiveInTravis);
      }
      throw err;
    });
```

As a side-effect, this also avoids calling _travisIsSyncing_ immediately after requesting a sync (or finding out that Travis is already syncing), which seems worthwhile, since it's highly likely that Travis will still be syncing then.

marco-c, line 360: I've made travisAwaitSyncing unconditionally wait five seconds.
I haven't added the error message check, even if it would avoid some useless calls, because I couldn't assess if it was reliable enough (it isn't cited anywhere in the official documentation and it's never used in the official libraries for JavaScript and Ruby, moreover it isn't easily reproducible, so regressions would go unnoticed) and because this is such a rare case that a couple more calls shouldn't really matter.

mykmelez, line 360: I'm more concerned about the complexity of the logic, which makes it harder to hack and can hide bugs. I'd prefer it to be as simple as possible (although no simpler). But I'm ok with the current version.

PR #85, opened 2015-10-15
Use yargs instead of gulp.env

marco-c:


gulpfile.js
```diff
@@ -54,7 +55,7 @@ gulp.task('serve', function () {
 });
 
 gulp.task('test', function () {
-  return gulp.src(gulp.env.file ? gulp.env.file : 'test/test*.js', {read: false})
+  return gulp.src(argv.file ? argv.file : 'test/test*.js', {read: false})
```
digitarald, line 58: - `file` as argument isn't very descriptive. `test` or `suite` might fit its single use case better.
- use `yargs.option` to set a default for file, like `const argv = yargs.option('s', {alias: 'suite', default: 'test/test*.js'})`

PR #91, opened 2015-10-15
specify repo URL to gh-pages to fix Travis auto-deploy

mykmelez:
This should work, although I haven't tested it yet, since I can't do so easily.


mykmelez:
I pushed a patch update (0.9.1) to npmjs.com with this change and then made https://github.com/mykmelez/test-app-5 use it to auto-deploy, which succeeded: https://travis-ci.org/mykmelez/test-app-5/builds/85642067. So this change does indeed fix the regression.

Issue #38, opened 2015-10-08
release template files to public domain


mykmelez:
We should release the template files to the public domain, so web developers who use them to bootstrap an app can choose their own license for the app. In some cases, this will require replacing template code that we obtained from other sources (like the GitHub Pages generator). The headers for files in the public domain are at the bottom of this page: https://www.mozilla.org/en-US/MPL/headers/.

PR #96, opened 2015-10-16
[WIP] Offliner integration

delapuente:
@mykmelez do you want to take a look, please?


lib/bootstrap.js
```diff
@@ -16,6 +16,8 @@
 
 'use strict';
 
+var Promise = require("es6-promise").Promise;
```
mykmelez: Since we require Node 0.12 now, Promise should be available already, right?

lib/bootstrap.js
```diff
@@ -110,7 +112,7 @@ function getDefaultTemplateConfig(dir) {
     name: "oghliner-template-app",
     repository: "https://oghliner-template-app.git",
     description: "A template app bootstrapped with oghliner.",
-    license: "Apache-2.0",
+    license: "Apache-2.0"
```
mykmelez: Nit: trailing commas in object literals are fairly conventional in this codebase, so I would leave this one.

lib/offline.js
```diff
@@ -42,7 +45,7 @@ module.exports = promisify(function(config, callback) {
   // Remove the existing service worker, if any, so sw-precache doesn't include
```
mykmelez: Nit: update comments that reference _sw-precache_, like this one, to reference _offliner_ instead.

lib/offline.js
```diff
@@ -58,14 +61,48 @@ module.exports = promisify(function(config, callback) {
...
-      staticFileGlobs: fileGlobs.map(function(v) { return path.join(rootDir, v) }),
-      stripPrefix: rootDir,
-      verbose: true,
-      logger: gutil.log,
-      importScripts: config.importScripts || [],
-      cacheId: cacheId,
-      ignoreUrlParametersMatching: config.ignoreUrlParametersMatching || [/./],
-    }, callback);
+    var absoluteGlobs =
+      fileGlobs.map(function (v) { return path.join(rootDir, v); });
```
mykmelez, line 65: Note that there's no line width restriction for this codebase. Statements like this are welcome to live on one line and exceed 80 columns:

``` js
    var absoluteGlobs = fileGlobs.map(function (v) { return path.join(rootDir, v); });
```

(That said, it's also ok to wrap lines; do whatever feels most readable!)

templates/app/index.html
```diff
@@ -6,7 +6,8 @@
     <title><%= name %></title>
     <meta name="description" content="<%= name %> : <%= description %>">
     <link rel="stylesheet" type="text/css" media="screen" href="styles/stylesheet.css">
-    <script src="scripts/offline-manager.js"></script>
+    <script data-worker="offliner-worker.js" src="lib/offliner/dist/offliner-client.min.js"></script>
+    <script defer src="scripts/index.js"></script>
```
mykmelez: The structure of the &lt;script> tag that enables offlining is a key part of the interface design of Oghliner, especially since its addition to the page(s)/template(s) of a pre-existing app is one of the only parts of the _integrate_ process that is not fully automated. So it's important that it be as simple and straightforward as possible.

Thus there are a few issues here:
1. Renaming the "manager" script to index.js is dangerous, because that name could conflict with another script in a pre-existing app. The original name, or another name that is less likely to conflict, would be better. (Of course this isn't a problem in the template app, but it'd be simpler to keep the name identical regardless of whether the file is being generated by _bootstrap_ or _integrate_.).
2. Renaming the worker script to offliner-worker.js is ok, but the reason I called it offline-worker.js instead of oghliner-worker.js or sw-precache-worker.js is that I didn't want to expose the implementation detail of which node module is being used to generate the code, since that detail is irrelevant to a developer who is looking at a script reference and wondering what the script does.
3. Requiring a developer to add and maintain two script tags instead of one is double the effort, and it's additional effort when one of them has a complicated _src_ path and an additional, nonstandard _data-worker_ attribute. Is it possible to integrate the functionality provided by offliner-client.min.js into the "manager" script, such the developer only has to integrate and maintain a single, simple script reference? It's fine to call it a "client" rather than a "manager", if that's more accurate; the key is just to simplify the code to reduce the developer burden to the absolute minimum. The ideal interface here is: 

``` html
<script src="offline-client.js"></script>
```

(In practice, we end up putting the file into a "scripts/" subdirectory, which is more conventional for web software projects.)

templates/app/lib/offliner/dist/offliner-client.min.js
```diff
@@ -0,0 +1 @@
...t){var n=this._eventListeners[e]||[];n.forEach(function(e){var n=e[0],r=e[1];n.call(r,t)})},_installMessageHandlers:function i(){function e(e){var n=e.data,r=n?n.type:"",s=r.split(":");"offliner"===s[0]&&t._handleMessage(s[1],n)}var t=this;if(!i.done){if("function"==typeof BroadcastChannel){var n=new BroadcastChannel("offliner-channel");n.onmessage=e}else navigator.serviceWorker.addEventListener("message",e);i.done=!0}},_handleMessage:function(e,t){navigator.serviceWorker;"xpromise"===e?this._resolveCrossPromise(t):this._runListeners(e,t)},_has:function(e,t,n){for(var r,s=this._eventListeners[e]||[],i=0;r=s[i];i++)if(r[0]===t&&r[1]===n)return!0;return!1},_xpromise:function(e){return new Promise(function(n,r){function s(e){r(new Error(e))}var i=t++,o={type:"xpromise",id:i,order:e};this._xpromises[i]=[n,s],this._send(o)}.bind(this))},_send:function(e){navigator.serviceWorker.getRegistration().then(function(t){return t&&t.active?t.active.postMessage(e):void warn("Not service worker active right now.")})},_resolveCrossPromise:function(e){var t=this._xpromises[e.id];t?t["rejected"===e.status?1:0](e.value):console.warn("Trying to resolve unexistent promise:",e.id)}}}(this.exports||this);
```
mykmelez: I'm not sure how scripts are typically minimized in gulp build flows, but I suspect that it'd be better to include a non-minimized version of this script in the template and then let the developer minimize the script (and others) at build time.

lib/bootstrap.js
delapuente: Right, confirmed with @brendandahl

templates/app/index.html
delapuente: @mykmelez, answering each item:
1. You're right. I will recover the original name.
2. I think is better to hide the details. I will rename it to the original name as well.
3. Well this is tricky. offliner is a library and it provides a way to communicate with the worker you design (in this case, the worker oghliner automatically writes). This is the client. [It is intended to be controlled by client code](https://github.com/mozilla/oghliner/pull/96/files#diff-6d5cc80376a3a8d696a6bd33cbbc2a9eR25), this is why we use two scripts: one is the library, the other is your code. I have two proposals to solve this:
   1. Add a special `data-autoinstall="true"` to offliner to make it install the worker automatically.
   2. Lazy-load the library in the client code before installing.

What do you think?

mykmelez: > Add a special data-autoinstall="true" to offliner to make it install the worker automatically.
> Lazy-load the library in the client code before installing.

I suggest the latter: lazy-load the library in the client. Then you don't have to change the behavior of _offliner_ right now, and the client can conditionally load the library only if `'serviceWorker' in navigator`, which'll save a network request on clients that don't support it.

delapuente:
Hi @mykmelez , can you give me a second opinion about the changes?

digitarald:
@marco-c could you check this while @mykmelez is out?

lib/offline.js
```diff
@@ -39,10 +42,10 @@ module.exports = promisify(function(config, callback) {
 
   var fileGlobs = config.fileGlobs || ['**/*'];
 
-  // Remove the existing service worker, if any, so sw-precache doesn't include
+  // Remove the existing service worker, if any, so offliner doesn't include
   // it in the list of files to cache.
   try {
-    fs.unlinkSync(path.join(rootDir, 'offline-worker.js'));
+    fs.unlinkSync(path.join(rootDir, 'offliner-worker.js'));
```
mykmelez: This change should also be reverted.

lib/offline.js
```diff
@@ -58,14 +61,48 @@ module.exports = promisify(function(config, callback) {
...
+      return matchings;
+    }, {}));
+  }
+
+  function getFilesAndHashes(files, stripPrefix) {
+    return files
+      .map(function (filepath) {
+        var data = fs.readFileSync(filepath);
+        var hash = getHash(data);
+        return { filepath: filepath.replace(stripPrefix, ''), hash: hash };
```
mykmelez: This will strip the prefix elsewhere in the string, if it doesn't appear at the beginning. That's unlikely, but in the event that it happens, we should avoid it, which you could do with a function second parameter that checks the offset of the string to replace:

```
filepath.replace(stripPrefix, function(v, offset) { if (offset === 0) return ''; });
```

It might also make sense to warn if the prefix isn't at the beginning of the path, unsure.

templates/app/scripts/offliner/safeguard.js
```diff
@@ -0,0 +1,5 @@
+/**
+ * This script exists for your convenience. Requests to this URL are never
+ * intercepted by the generated service worker. It is here to allow you
+ * modify the behaviour of your app even when you cached unintended content.
+ */
```
mykmelez: Can you explain the use case(s) for this? It seems orthogonal to the purpose of this pull request. Perhaps we can consider it in a separate PR?

lib/offline.js
mykmelez: (I'm aware that sw-precache also has this problem. It's also quite the edge case. Still, I'd fix it.)

lib/offline.js
```diff
@@ -58,14 +61,48 @@ module.exports = promisify(function(config, callback) {
...
+      })
+      .map(function (pathAndHash) {
+        return '  "' + pathAndHash.filepath + '" /* ' + pathAndHash.hash + ' */';
+      })
+      .join(',\n');
```
mykmelez: Instead of joining the array into a string here, can we pass the array itself to the template and have the template iterate over it? That would preserve the structure of the data for longer, which is generally safer.

lib/offline.js
delapuente: I'll do.

lib/offline.js
delapuente: Sure.

templates/app/scripts/offliner/safeguard.js
delapuente: Yes, I think it is better to add a separated issue.

delapuente:
Hi @mykmelez , what about another round, please?

lib/offline.js
```diff
@@ -59,29 +61,49 @@ module.exports = promisify(function(config, callback) {
...
-        var stat = fs.statSync(file);
+    stream.on('finish', function () { callback(); });
+    stream.on('error', function (e) { callback(e); });
+  });
 
-        if (stat.isFile() && stat.size > 2 * 1024 * 1024 && staticFileGlobs.indexOf(file) === -1) {
-          gutil.log(gutil.colors.red.bold(file + ' is bigger than 2 MiB. Are you sure you want to cache it? To suppress this warning, explicitly include the file in the fileGlobs list.'));
-        }
+  function flatGlobs(fileGlobs) {
+    return Object.keys(fileGlobs.reduce(function (matchings, fileGlob) {
```
mykmelez, line 80: Nit: matchings -> matches

mykmelez:
Two other issues:
1. The _integrate_ command copies scripts/offline-manager.js into the app and tells the user to load it: https://github.com/mozilla/oghliner/blob/master/lib/integrate.js#L41-L53. But on this branch it would also need to copy all the files in the scripts/offliner/ subdirectory.
2. _offliner_ is licensed under an Apache 2.0 license, but in #38 we plan to release all template files to the public domain, so we would need to be able to release the _offliner_ files in the templates/ directory to the public domain as well.

delapuente:
After talking with @mykmelez, we decided to close this issue without merging and I will be writing the service worker we need to address #154 and #5.

## 

Apart from the legal issue of making all of public domain, which is a minor problem, the main reason is the functionality covered by _offliner_ is beyond the scope of _oghliner_ **v1**. Both _offliner_ and _oghliner_ are trying to provide an application lifecycle from different approaches and both need fine-grain control on service workers. My intention integrating _offliner_ was to provide flexibility while avoiding rethink the application lifecycle but for _oghliner v1_ we want to experiment with a simpler solution more tied to the service worker lifecycle.

Thus, to regain full control on the lifecycle, we are writing our own service worker which will provide the precise functionality we want.

PR #99, opened 2015-10-16
Catch fatal error in the case that the repository has no commits

darkwing:
If you attempt to `deploy` on a repository with no commits, you receive a fatal error directly from `childProcess.execSync`.  This pull requests mentions that in documentation but also programmatically catches said error and provides a more elegant notification for the user.

First commit, be gentle.  :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,6 +34,12 @@ Invoke `gulp` to rebuild your app and regenerate the script that offlines it. In
 gulp && gulp deploy
 ```
 
+At least one commit to the repository is required for successful deploy.  The following could be used to commit the changes by Oghliner to the respository:
+
+```bash
+git add . && git commit -m "Initial version of app"
+```
+
 Using The Tool
 --------------
 
--- a/lib/deploy.js
+++ b/lib/deploy.js
@@ -32,7 +32,19 @@ module.exports = promisify(function(config, callback) {
   var ghPagesConfig = {};
   var rootDir = config.rootDir ? config.rootDir : '.';
   var cloneDir = config.cloneDir ? config.cloneDir : null;
-  var commitMessage = config.message ? config.message : childProcess.execSync('git log --format=%B -n 1').toString().trim();
+
+  var commitMessage = config.message || (function() {
+    var spawn = childProcess.spawnSync('git', ['log', '--format=%B', '-n', '1']);
+    var errorText = spawn.stderr.toString().trim();
+
+    if (errorText) {
+      gutil.log('Fatal error from `git log`.  You must have one commit before deploying.');
+      throw new Error(errorText);
+    }
+    else {
+      return spawn.stdout.toString().trim();
+    }
+  })();
 
   gutil.log('Deploying "' + commitMessage + '"');
 


README.md
```diff
@@ -34,6 +34,8 @@ Invoke `gulp` to rebuild your app and regenerate the script that offlines it. In
 gulp && gulp deploy
 ```
 
+*At least one commit to the repository is required for successful deploy.*
```
mykmelez: Nice! We might want to suggest that the developer can commit the changes that the bootstrap command just made, and give them a sample command to do it, something like:

``` bash
git add . && git commit -m"initial version of app"
```

But we can also wordsmith that in a followup.

lib/deploy.js
```diff
@@ -32,7 +32,19 @@ module.exports = promisify(function(config, callback) {
   var ghPagesConfig = {};
   var rootDir = config.rootDir ? config.rootDir : '.';
   var cloneDir = config.cloneDir ? config.cloneDir : null;
-  var commitMessage = config.message ? config.message : childProcess.execSync('git log --format=%B -n 1').toString().trim();
+
+  var commitMessage = config.message || (function() {
+    var spawn = childProcess.spawnSync('git', ['log', '--format=%B', '-n', '1']);
+    var errorText = spawn.stderr.toString();
```
mykmelez: Nit: It looks like this has a trailing newline character attached to it, so it'd be helpful to `.trim()` it.

lib/deploy.js
```diff
@@ -32,7 +32,19 @@ module.exports = promisify(function(config, callback) {
...
+
+    if (errorText) {
+      gutil.log('Fatal error from `git log`.  You must have one commit before deploying.  Error detail: ' + errorText);
+      throw new Error();
```
mykmelez: Nit: I'd make the raw _errorText_ be the Error object's _message_ instead of appending it to the friendly message, i.e. instantiate the Error object with `new Error(errorText)`.

Then, instead of seeing:

> [18:15:23] Fatal error from `git log`.  You must have one commit before deploying.  Error detail: fatal: bad default revision 'HEAD'
> [Error]

A developer will see this:

> [18:17:41] Fatal error from `git log`.  You must have one commit before deploying.
> [Error: fatal: bad default revision 'HEAD']

Which feels a bit more readable.

darkwing: Note to self for Monday:  it looks like other places in the deploy script do a `callback` call for the error.

lib/deploy.js
```diff
@@ -32,7 +32,20 @@ module.exports = promisify(function(config, callback) {
...
-  var commitMessage = config.message ? config.message : childProcess.execSync('git log --format=%B -n 1').toString().trim();
+
+  var commitMessage = config.message || (function() {
+    var spawn = childProcess.spawnSync('git', ['log', '--format=%B', '-n', '1']);
+    var errorText = spawn.stderr.toString().trim();
+
+    if (errorText) {
+      gutil.log('Fatal error from `git log`.  You must have one commit before deploying.');
+      throw new Error(errorText);
+      callback(spawn.stderr);
```
darkwing: Let me know what you think of this; I see in other places that we're calling `callback` when an error occurs, so I've added that as well.

marco-c: _promisify_ should do the right thing both if you throw an error and if you call the callback with the error argument (which is the first argument), so it shouldn't really matter.
Maybe for consistency with the rest of the function, you could call the callback. Note, however, that you're currently calling it after throw, which means it'll never be called. You could instead do:

``` JavaScript
if (errorText) {
  gutil.log('Fatal error from `git log`.  You must have one commit before deploying.');
  callback(spawn.stderr);
  return;
}
```

darkwing: So executing the callback doesn't halt the gulp process and results in uglier messages down the road.  For the sake of usability, I think we should use `new Error()` and just halt the process completely.  Thoughts?

marco-c: Ah, I didn't notice this was called inside another function, so the return isn't affecting the enclosing function.

You could use _throw new Error_ then and modify the gulp script from:

``` JavaScript
gulp.task('deploy', function(callback) {
  oghliner.deploy({
    rootDir: 'dist',
  }, callback);
});
```

to:

``` JavaScript
gulp.task('deploy', function() {
  return oghliner.deploy({
    rootDir: 'dist',
  });
});
```

This way the gulp task halts when the error is thrown because the promise returned by deploy gets rejected.

Or you could remove the internal function, write everything in the enclosing function and use `callback(...) return;`.

I'd prefer the second option (or any other option where the callback is called and the gulp task halted) because I think in the future we can avoid using _promisify_ and directly return a promise here, so a call to _callback_ could be simply replaced by a call to _reject_.

marco-c: (Actually I think we should modify the gulp tasks to return promises anyway, so I've opened #102).

darkwing:
Looks like I broke test...will update

lib/deploy.js
marco-c: Here's the difference in the gulp output, if you use _throw_, between the two versions I posted above:

```
[18:01:09] Using gulpfile ~/Documents/workspace/oghliner/asd/gulpfile.js
[18:01:09] Starting 'deploy'...
[18:01:09] Fatal error from `git log`.  You must have one commit before deploying.
```

```
[18:01:48] Using gulpfile ~/Documents/workspace/oghliner/asd/gulpfile.js
[18:01:48] Starting 'deploy'...
[18:01:48] Fatal error from `git log`.  You must have one commit before deploying.
[18:01:48] 'deploy' errored after 8.57 ms
[18:01:48] Error: fatal: Not a git repository (or any parent up to mount point /home)
Stopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).

    at formatError (/usr/local/lib/node_modules/gulp/bin/gulp.js:169:10)
    at Gulp.<anonymous> (/usr/local/lib/node_modules/gulp/bin/gulp.js:195:15)
    at Gulp.emit (events.js:107:17)
    at Gulp.Orchestrator._emitTaskDone (/home/marco/Documents/workspace/oghliner/asd/node_modules/gulp/node_modules/orchestrator/index.js:264:8)
    at /home/marco/Documents/workspace/oghliner/asd/node_modules/gulp/node_modules/orchestrator/index.js:275:23
    at finish (/home/marco/Documents/workspace/oghliner/asd/node_modules/gulp/node_modules/orchestrator/lib/runTask.js:21:8)
    at /home/marco/Documents/workspace/oghliner/asd/node_modules/gulp/node_modules/orchestrator/lib/runTask.js:45:4
    at tryCallOne (/home/marco/Documents/workspace/oghliner/oghliner/node_modules/promisify-node/node_modules/nodegit-promise/lib/core.js:37:12)
    at /home/marco/Documents/workspace/oghliner/oghliner/node_modules/promisify-node/node_modules/nodegit-promise/lib/core.js:103:15
    at flush (/home/marco/Documents/workspace/oghliner/oghliner/node_modules/promisify-node/node_modules/nodegit-promise/node_modules/asap/raw.js:50:29)
```

darkwing:
OK, updated to terminate the process if no initial commit is there.

README.md
```diff
@@ -34,6 +34,12 @@ Invoke `gulp` to rebuild your app and regenerate the script that offlines it. In
 gulp && gulp deploy
 ```
 
+At least one commit to the repository is required for successful deploy.  The following could be used to commit the changes by Oghliner to the respository:
```
mykmelez, line 37: Nit: respository -> repository

Otherwise looks great!

@marco-c You also reviewed this, can you merge it if you're happy with it?

PR #105, opened 2015-10-20
Add live test for Oghliner used as a tool with CLI commands

marco-c:


marco-c:
Once I get this working, there are still a couple of outstanding problems:
- What to do when the test is executed locally?
- What to do when the test is executed in a PR? (which, unless the PR comes from a branch in this repository, doesn't have access to the secure Travis env variables)

marco-c:
It is working now. Any idea on how to address those problems?

marco-c:
For now with cac4567, we're skipping the _live_ tests when the USER or PASS environment variables aren't set (e.g. pull requests).

test/testLiveAsTool.js
```diff
@@ -0,0 +1,150 @@
...
+
+  afterEach(function() {
+    process.chdir(oldWD);
+
+    delete process.env['GH_TOKEN'];
+  });
+
+  it('should work', function() {
+    return createRepo()
+    .then(() => spawn('git', ['clone', 'https://' + username + ':' + password + '@github.com/' + username + '/test_oghliner_live']))
```
mykmelez: Arrow functions aren't supported in Node 0.12, with which we still claim compatibility.

mykmelez: Erm, I just noticed the _--harmony_ flag you added to the _mocha_ invocation.

marco-c: Yeah, I'm just using them for tests. It could (unlikely) cause problems, because if we use ES6 features in the library the tests would still pass (but we don't want to require running _oghliner_ with _--harmony_).

mykmelez: They cause tests to fail if you run them via `mocha` or `gulp test`.

mykmelez:
> Once I get this working, there are still a couple of outstanding problems:
> 
> ```
> What to do when the test is executed locally?
> What to do when the test is executed in a PR? (which, unless the PR comes from a branch in this repository, doesn't have access to the secure Travis env variables)
> ```

If the goal is to test the CLI, then the tests don't need access to GitHub for that, since the _git_ command works identically when a remote repository is actually a local directory. So the tests could create a local directory to serve as the "remote," `git init` in it, then `git clone` that remote to another local directory that serves as the local clone and run the tests in that.

marco-c:
> If the goal is to test the CLI, then the tests don't need access to GitHub for that, since the git command works identically when a remote repository is actually a local directory. So the tests could create a local directory to serve as the "remote," git init in it, then git clone that remote to another local directory that serves as the local clone and run the tests in that.

The goal was to also test the integration with GitHub, but actually it isn't strictly needed for this test (it would be more useful for the test where there's also _configure_, which uses the GitHub API).

marco-c:
I'll test _configure_ as well here. Travis is currently stuck syncing the repositories of the testing user I've created, hopefully the issue will be fixed tomorrow.

mykmelez:
> I'll test configure as well here. Travis is currently stuck syncing the repositories of the testing user I've created, hopefully the issue will be fixed tomorrow.

@wfwalker had that problem with his normal user account, so he tweeted at Travis, and they said it'll take a while because he's a member of the Mozilla organization on GitHub: https://twitter.com/wfwalker/status/650050216452685824

marco-c:
> @wfwalker had that problem with his normal user account, so he tweeted at Travis, and they said it'll take a while because he's a member of the Mozilla organization on GitHub: https://twitter.com/wfwalker/status/650050216452685824

The testing account has a single repository and isn't a member of any organization, so hopefully it won't take too much. Unfortunately I can't log-in today (the script is stuck at `Getting Travis token…` and the Travis website keeps loading indefinitely when I hit the 'Sign in' button).

After a while, the script fails with `{ [Error: socket hang up] code: 'ECONNRESET' }`

wfwalker:
The twitter folks told me I was having problems because they were busy adding all possible Mozilla organization repos to my Travis account. It eventually started working as expected (after an hour or two, maybe?)

mykmelez:
Would it be possible to add support for specifying a 2FA code? When I test this with my normal GitHub account, which has 2FA enabled, I fail on:

```
  1) CLI interface, oghliner as a tool should work:
     {"message":"Must specify two-factor authentication OTP code.","documentation_url":"https://developer.github.com/v3/auth#working-with-two-factor-authentication"}
```

marco-c:
> Would it be possible to add support for specifying a 2FA code?

Unfortunately it wouldn't be really easy to automate it. If you specify your OTP code when executing the test, it will work at the beginning but eventually expire before the test finishes.

mykmelez:
> Unfortunately it wouldn't be really easy to automate it. If you specify your OTP code when executing the test, it will work at the beginning but eventually expire before the test finishes.

Right, the tests would have to finish within the time period of the code's validity. I read somewhere that GitHub accepts the both the current and previous code, effectively doubling the validity period.

Alternately, can the test be made to prompt for an auth code when needed?

test/testLiveAsTool.js
```diff
@@ -0,0 +1,300 @@
...
+      useOTP = true;
+      return createAuthorization();
+    }
+
+    if (error.message === 'Validation Failed' && error.errors[0].code === 'already_exists') {
+      return getTokenId().then(deleteAuthorization).then(createAuthorization);
+    }
+
+    throw err;
+  })
```
mykmelez: Nit: add semi-colon (for consistency with usage elsewhere in this file).

mykmelez:
The first time I ran the current version of the test, it prompted me for an auth code once. But on subsequent runs, it has prompted me to enter an auth code four times in a row before showing any other output.

This appears to be unrelated to the time remaining for the code to be valid. I see the problem whether I'm entering a code that is about to expire, one that is newly valid, or one in the middle of its validity period, and regardless of whether or not the code is regenerated in the middle of the sequence, as it is in this case:

>  CLI interface, oghliner as a tool
> Auth Code: 528903
> Auth Code: 528903
> Auth Code: 528903
> Auth Code: 795615
> Cloning into 'test_oghliner_live'...

Each prompt does seem to require a valid code: if I enter an invalid code four times, I get a fifth prompt, and if I then enter a valid code, I get another prompt, etc. up to four times, as happened here:

>  CLI interface, oghliner as a tool
> Auth Code: 111111
> Auth Code: 111111
> Auth Code: 111111
> Auth Code: 111111
> Auth Code: 621018
> Auth Code: 621018
> Auth Code: 621018
> Auth Code: 621018
> Cloning into 'test_oghliner_live'...

Later on in the test run, after my current code has expired, I'm prompted again, but only once:

> You're using two-factor authentication with GitHub.
> Please enter the code provided by your authentication software.
> 
> Auth Code: 255341  
> 
> You already have the GitHub token "Oghliner token for mykmelez/test_oghliner_live".

I haven't dug into the code yet, but perhaps it isn't correctly caching and reusing the code I enter initially, such that it re-prompts me each time it makes a request that fails due to the need to include a code?

marco-c:
> I haven't dug into the code yet, but perhaps it isn't correctly caching and reusing the code I enter initially, such that it re-prompts me each time it makes a request that fails due to the need to include a code?

Yes, I haven't implemented any caching to avoid complicating the code. There's at least one place where caching wouldn't be possible (when the `configure` command is executed), so I haven't implemented it altogether but I'm always re-asking it.

mykmelez:
> Yes, I haven't implemented any caching to avoid complicating the code. There's at least one place where caching wouldn't be possible (when the configure command is executed), so I haven't implemented it altogether but I'm always re-asking it.

That explains it. ;-)

Ok, I'm fine with landing this as-is, but here are the followups I'd like to see us file and track:
- [x] prompt user for username/password (will require a different mechanism for enabling the test), so it isn't necessary to set those in environment variables, which exposes some credential-revealing footguns
- [x] cache and reuse 2FA OTP (auth) code
- [x] option to delete test repository (and resync Travis) after test run completes, i.e. clean up after test, so it's possible to re-run the test from scratch repeatedly without cumbersome manual cleanup

marco-c:
> Ok, I'm fine with landing this as-is, but here are the followups I'd like to see us file and track:

Filed #137, #138 and #139.

PR #106, opened 2015-10-20
Use promisified-promptly

marco-c:
There are a few issues in promisifying the `promptly` module, so I've built https://github.com/marco-c/node-promisifed-promptly to promisify it manually.

This PR also removes the dependency on `inquirer` by using promptly in lib/bootstrap.js as well.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,6 @@ before_script:
   - 'git config --global user.email "myk@mykzilla.org"'
 script:
   - gulp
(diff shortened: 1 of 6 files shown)


marco-c:
There's a race condition which is making the test fail, 'Password:' is printed before the call to await('Password:').

marco-c:
Looks like the race condition was the cause of the failure with `gulp test`, so I've re-enabled running tests with `gulp test`.

PR #107, opened 2015-10-20
unfold deploy command (fixes #19)

### LINKED ISSUE #19: make Travis show output of deploy command by default
The output of the command that _configure_ adds to the _after_success_ script in .travis.yml is collapsed by default in Travis build logs. So you only see a single line like this when you first view a log ([like this one](https://travis-ci.org/mykmelez/mc-choir/builds/83798486)):

> 2.87s$ [ "${TRAVIS_PULL_REQUEST}" = "false" ] && [ "${TRAVIS_BRANCH}" = "master" ] && gulp deploy

You can see the rest of the log by clicking on the disclosure widget (expander arrow) to the left of that line item, after which you see the output of the command, which might look something like this:

> [23:31:35] Using gulpfile ~/build/mykmelez/mc-choir/gulpfile.js
> [23:31:35] Starting 'deploy'...
> [23:31:37] Finished 'deploy' after 1.91 s

But it would be better if Travis showed you that output by default, so you didn't have to expand the command to see it.

mykmelez:
Per https://github.com/travis-ci/travis-ci/issues/2285, I think this'll fix #19, although I'm unsure how to test it without actually merging this change.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,4 +12,4 @@ script:
   - gulp
   - node node_modules/mocha/bin/mocha test
 after_success:
-  - '[ "${TRAVIS_PULL_REQUEST}" = "false" ] && [ "${TRAVIS_BRANCH}" = "master" ] && gulp deploy'
+  - 'echo "travis_fold:end:after_success" && [ "${TRAVIS_PULL_REQUEST}" = "false" ] && [ "${TRAVIS_BRANCH}" = "master" ] && echo "Deploying…" && gulp deploy'


marco-c:
You could try to add another command (which runs also on PRs) to after_success and see if it is unfolded with your change.

mykmelez:
> You could try to add another command (which runs also on PRs) to after_success and see if it is unfolded with your change.

Sadly, adding another command to the script violates the Travis Uncertainty Principle, causing the identity of the first command to change from `travis_fold:end:after_success` to `travis_fold:end:after_success.1`. And it triggers a fold marker for the second command, which I would then have to disable.

A simpler test is to modify the existing command to deploy unconditionally, which I've done in 2bffb1c. It's build https://travis-ci.org/mozilla/oghliner/builds/86559661 shows that the command is correctly unfolded, and the long, complex command is replaced by the clear string `Deploying…`.

Of course the deploy fails with an authentication failure, since it tries to deploy a branch in another repository, and Travis doesn't export the GitHub token environment variable in that case. But the point is just to test that the output of _deploy_ is not folded, and it isn't. Now I'll undo that change, squash the commits on this branch, and remove the WIP designation.

Note that there's still one folded line, because Travis writes a `travis_fold:start:after_success` fold marker to the log before the command writes `travis_fold:end:after_success`. But that's a minor issue, not worth worrying about.

Here's a screenshot of the build output:

screen shot 2015-10-21 at 00 02 27

And here's the raw log:

```
[32;1mThe command "node node_modules/mocha/bin/mocha test" exited with 0.[0m
travis_fold:start:after_success
[0Ktravis_time:start:0a34c09c
[0K$ echo "travis_fold:end:after_success" && echo "Deploying…" && gulp deploy
travis_fold:end:after_success
Deploying…
```

mykmelez:
I'm unsure why GitHub doesn't reference the merge commit build, but it's https://travis-ci.org/mozilla/oghliner/builds/86594069, and it confirms that the _deploy_ command's output is no longer folded:

screen shot 2015-10-21 at 08 00 03

marco-c:
:+1:

PR #110, opened 2015-10-21
Use template offline-manager.js file for integrate command.

brendandahl:
The packaged node module does not include app/scripts/offline-manager.js

As discussed on IRC with Marco, we should really be doing a live test or test the packaged node module.

### DIFF
--- a/test/testIntegrate.js
+++ b/test/testIntegrate.js
@@ -27,34 +27,13 @@ describe('Integrate', function() {
     });
   });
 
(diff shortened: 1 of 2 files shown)


test/testIntegrate.js
```diff
@@ -27,34 +27,13 @@ describe('Integrate', function() {
     });
   });
 
-  it('should throw when the offline-manager.js script doesn\'t exist', function() {
```
brendandahl, line 30: Other tests should fail if this is not true.

Issue #113, opened 2015-10-21
`configure` CLI command doesn't terminate when invoked via process.spawn


marco-c:
This could be related to #112.
Replacing:

```
configure()
.catch(function(err) {
  console.error(err);
});
```

with:

```
configure()
.catch(function(err) {
  console.error(err);
});
.then(process.exit);
```

fixes the problem for me, so this is probably what I'm going to do as part of #105, but we should find the underlying issue.

PR #114, opened 2015-10-22
Use setTimeout to fix race with prompt.

### LINKED ISSUE #112: Tests intermittently hang at the end
After `Finished 'test' after 1.88 s`.

brendandahl:
Fixes #112 for me. I still was not able to fully track down where this issue is coming from. Everything seems okay on our side so I'm thinking it's in the read module, but I'm not sure it's worth spending more time on it.

### DIFF
--- a/test/testConfigure.js
+++ b/test/testConfigure.js
@@ -56,7 +56,10 @@ describe('Configure', function() {
   }
 
   function emit(data) {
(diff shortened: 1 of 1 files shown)


marco-c:
Looks good to me as a temporary solution. I'd still investigate this failure as #113 (which is a real world issue) could be related.

PR #118, opened 2015-10-24
Test 'configure' when a .travis.yml file already exists

marco-c:


test/testConfigure.js
```diff
@@ -635,6 +638,128 @@ describe('Configure', function() {
...
+    return enterUsernamePassword()
+    .then(function() {
+      return await('You\'re ready to auto-deploy using Travis!');
+    })
+    .then(function() {
+      var travisYml = readYaml.sync('.travis.yml');
+      expect(travisYml.after_success).to.have.length(2);
+      expect(travisYml.after_success[0]).to.equal('a_command');
+      expect(travisYml.after_success[1]).to.equal(
+        'echo "travis_fold:end:after_success" && ' +
```
brendandahl: This is testing what a configuration setting is, which is not the scope of this test. Can we make it a more general assertion?

test/testConfigure.js
```diff
@@ -635,6 +638,128 @@ describe('Configure', function() {
...
+      return await('You\'re ready to auto-deploy using Travis!');
+    })
+    .then(function() {
+      var travisYml = readYaml.sync('.travis.yml');
+      expect(travisYml).to.include.keys('env');
+      expect(travisYml.env).to.include.keys('global');
+      expect(travisYml.env).to.include.keys('matrix');
+      expect(travisYml.env.global).to.have.length(2);
+      expect(travisYml.env.global[0]).to.equal('ENV_GLOBAL');
+      expect(travisYml.env.global[1]).to.have.keys('secure');
```
brendandahl: Again testing configuration.

brendandahl:
Seems there a several spots that rely on the generated configuration, which should be orthogonal to these tests.

test/testConfigure.js
marco-c: The purpose of this check was to ensure we don't overwrite a custom command in _after_success_, but we still add the auto-deploy command.

test/testConfigure.js
marco-c: I added this check to make sure we don't overwrite the already existing env in .travis.yml, but we add the one we need.

test/testConfigure.js
brendandahl: Yes, but these can be written in a way that doesn't require the exact strings to be in the unit tests (either just checking the existence of an extra setting or programmatically reading in our travis file and comparing values). As with the other tests if we change the travis file we'll have to update all of these tests.

marco-c:
I made the tests more generic in d8c10af.

PR #119, opened 2015-10-25
Add live test for Oghliner used as a template with CLI commands

marco-c:


marco-c:
I need to update this to support 2FA like #105.

test/liveUtils.js
```diff
@@ -0,0 +1,208 @@
...
+      });
+
+      resolve(res);
+    });
+  }).catch(function(err) {
+    var error = JSON.parse(err.message);
+
+    if (error.message === 'Must specify two-factor authentication OTP code.') {
+      useOTP = true;
+      return createAuthorization(uesrname, password);
```
mykmelez: uesrname -> username

marco-c: Right, you made me remember that I haven't tested it yet with 2FA:

test/liveUtils.js
```diff
@@ -0,0 +1,272 @@
...
+var github = new GitHub({
+  version: '3.0.0',
+  protocol: 'https',
+  headers: {
+    'user-agent': 'Oghliner',
+  },
+});
+
+var liveUtils = {
+  repoName: 'test_oghliner_live' + process.version + process.pid,
```
mykmelez: Nit: these repo names would be a bit easier to read with separator characters between the version and id, f.e.: `test_oghliner_livev0.12.729326` -> `test_oghliner_live_v0.12_729326`.

mykmelez:
It would be helpful to make test cleanup optional, so you can test _configure_ with both a new repo and an existing one. That'd also require us to reuse repo names instead of generating a new one each time. We can do this in a followup

mykmelez:
Looks like this needs conflict resolution since I merged a couple other fixes.

marco-c:
> It would be helpful to make test cleanup optional, so you can test configure with both a new repo and an existing one. That'd also require us to reuse repo names instead of generating a new one each time. We can do this in a followup

Filed #164.

marco-c:
Merging this so I can update #159.

Issue #26, opened 2015-10-06
use formatting to make command output clearer


mykmelez:
Oghliner's output is quite simplistic, just plain text without any special formatting. By contrast, other popular command-line tools, like Yeoman (`npm install -g yo && yo`), use color, boldness, and other stylistic formatting to clarify the text they output to the console. (Yeoman even prints ASCII art, although that's a bit extraneous.)

We should evaluate the output of the various Oghliner commands for opportunities to use formatting to make them clearer.
- [x] Use gutil.log to log messages (#131)
- [x] Print colored warnings and errors (#131)
- [x] Clarify output of `configure` (#148)
- [x] Print large file warning in yellow (#151)
- [x] Print all suggestions in blue (#156)
- [x] Clarify output of `bootstrap` (#160)
- [x] Clarify output of `integrate` (#161)
- [x] Clarify output of `offline` (#162)
- [x] Clarify output of `deploy` (#163)

marco-c:
We should also improve error messages, using `console.error` doesn't give much details.

marco-c:
#131 improves the situation.

mykmelez:
It might be a bit further afield, but I see enhancements like #141 as keeping with the spirit of this issue.

mykmelez:
All dependencies resolved, so this too is resolved!

PR #122, opened 2015-10-26
Bold error messages, util naming

darkwing:
Simple update to make errors more noticeable.  https://github.com/mozilla/oghliner/issues/26


marco-c:
There are probably other messages which we want to format differently (e.g. questions), but this is a good start.

PR #123, opened 2015-10-27
Remove bootstrap's npmInstall option

marco-c:
This makes the test a little slower. Locally you can run tests with the `--skip-install` argument (e.g. `gulp test --skip-install`) to achieve the same effect of `npmInstall: false`.


marco-c:
We haven't updated the oghliner version when we started supporting Node.js v4, so this is failing when using Node.js v4.
Requires #124 (where I'm updating the version) in order to work.

PR #127, opened 2015-11-02
Cache large files by default

### LINKED ISSUE #28: cache large files with warning (and option for excluding them)
sw-precache sets its _maximumFileSizeToCacheInBytes_ to 2MiB by default, which breaks offlining of large image files, even if you explicitly include them via file globs. I understand the reasoning behind avoiding the caching of large files, especially if they aren't actually necessary for your app to function. But it does complicate and undermine the promise of Oghliner: that you can point it at your app's assets, and it'll ensure they're available offline.

Instead of simply excluding large files, we should warn the user when offlining them and give them the opportunity to exclude them from the cache, f.e. by allowing entries in _fileGlobs_ to exclude files instead of merely including files. Then a user who is warned about a large file can choose to cache it anyway by doing nothing or explicitly exclude it to prevent it from being cached, f.e. by using a leading exclamation point to negate the match à la [minimatch](https://github.com/isaacs/minimatch):

``` json
{
    "fileGlobs": [
        "!images/my-large-file.png",
    ]
}
```

We might even let users explicitly include a file to suppress the "large file" warning:

``` json
{
    "fileGlobs": [
        "images/my-large-file.png",
    ]
}
```

marco-c:
Fixes #28.


test/testOffline.js
```diff
@@ -149,4 +182,129 @@ describe('Offline', function() {
...
+      var content = fs.readFileSync(path.join(dir, 'offline-worker.js'), 'utf8');
+      assert.equal(content.indexOf('test_file_1.js'), -1);
+      assert.notEqual(content.indexOf('test_file_2.js'), -1);
+      assert.notEqual(content.indexOf('test_file_3.js'), -1);
+    });
+
+    return Promise.all([ checkWarnings, offlinePromise ]);
+  });
+
+  it('should not warn about explicitely included files', function() {
```
mykmelez: Nit: explicitely -> explicitly

lib/offline.js
```diff
@@ -58,14 +59,29 @@ module.exports = promisify(function(config, callback) {
...
+    var staticFileGlobs = fileGlobs.map(function(v) {
+      return path.join(rootDir, v)
+    });
+
+    staticFileGlobs.forEach(function(globPattern) {
+      glob.sync(globPattern.replace(path.sep, '/')).forEach(function(file) {
+        var stat = fs.statSync(file);
+
+        if (stat.isFile() && stat.size > 2 * 1024 * 1024 && staticFileGlobs.indexOf(file) === -1) {
+          gutil.log(gutil.colors.red.bold(file + ' is bigger than 2 MiB. Are you sure you want to cache it?'));
```
mykmelez: It'd also help to display a message about how to suppress these warnings, if any of them are displayed. It could say something like "To suppress 'file is bigger' warnings, explicitly include the file in the list of files to include." We can do this in a followup, however.

PR #128, opened 2015-11-02
Allow specifying files to deploy

marco-c:


lib/deploy.js
```diff
@@ -33,6 +34,20 @@ module.exports = promisify(function(config, callback) {
   var rootDir = config.rootDir ? config.rootDir : '.';
   var cloneDir = config.cloneDir ? config.cloneDir : null;
 
+  if (config.fileGlobs) {
+    fs.mkdirSync('tmp');
```
mykmelez: Can we use the _temp_ module here to create and (automatically) destroy a temp directory that is guaranteed not to already exist?

lib/integrate.js
```diff
@@ -19,27 +19,7 @@
...
-    srcStream.pipe(fs.createWriteStream(dest));
-
-    srcStream.on('error', reject);
-
-    srcStream.on('end', function() {
-      console.log(gutil.colors.blue(
-        'Your app needs to load the script offline-manager.js in order to register the service\n' +
-        'worker that offlines your app. To load the script, add this line to your app\'s HTML\n' +
-        'page(s)/template(s):\n\n' +
-        '\t<script src="' + dest + '"></script>'
```
mykmelez, line 36: We should still show this message, even if we aren't using this custom _copyFile_ method anymore.

lib/deploy.js
marco-c: It wasn't possible in the first implementation of my patch (because I was using the _src_ option, which accepts a relative directory), but it is possible now.
Done in 0f5c102.

lib/integrate.js
marco-c, line 36: Oops. Added back in 14bc55f.

PR #131, opened 2015-11-03
Use gutil.log to log messages; log errors/warnings in red

marco-c:


mykmelez:
I don't see any warnings being logged here, but shouldn't we log those in a different color?

Also, I'm not sure this quite fixes #26. It's certainly part of the solution, but #26 is a bit broader (and, ok, also a bit vaguer).

marco-c:
We're currently logging both errors and warnings in red. What color could we use for warnings?

mykmelez:
The most conventional pattern I've seen is to color warning messages _yellow_. See for example http://isabelcastillo.com/error-info-messages-css and http://www.jankoatwarpspeed.com/css-message-boxes-for-different-message-types/.

PR #135, opened 2015-11-03
copy Tabzilla files directly to dist/

mykmelez:
@darkwing What do you think about copying the Tabzilla files directly to the dist/ directory? Then we wouldn't have to .gitignore them, as the entire dist/ directory is .gitignored already. And old files would be less likely to stick around, since the dist/ directory is more likely to be blown away and recreated from scratch.

(Currently, old files aren't deleted from app/styles/tabzilla/, so if a Tabzilla update changes the files we copy, f.e. updating the name of an image file, then the old files will stick around in app/styles/tabzilla/. Copying directly to dist/ doesn't directly fix that, but it makes it easier to do and more likely to happen.)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,3 @@
 node_modules/
 dist/
 coverage/
-app/styles/tabzilla/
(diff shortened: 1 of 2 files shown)


darkwing:
Nice!  R+

PR #136, opened 2015-11-04
Directly use promises instead of promisifying 'deploy' and 'offline'

### LINKED ISSUE #129: Don't use promisify to promisify 'deploy' and 'offline', but directly use Promises

marco-c:
Fixes #129.


mykmelez:
Needs conflict resolution (from ESLint integration, I think).

PR #140, opened 2015-11-04
Add ESLint to tidy up

darkwing:
Feel free to tell me to go away... :dart:

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -20,6 +20,7 @@ var connect = require('gulp-connect');
 var gulp = require('gulp');
 var mocha = require('gulp-mocha');
 var istanbul = require('gulp-istanbul');
+var eslint = require('gulp-eslint');
 var argv = require('yargs').argv;
 var oghliner = require('./index.js');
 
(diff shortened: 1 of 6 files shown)


mykmelez:
I love tidiness!

gulpfile.js
```diff
@@ -67,9 +68,18 @@ gulp.task('pre-test', function () {
...
     .pipe(mocha())
     .pipe(istanbul.writeReports());
-})
+});
+
+gulp.task('lint', function() {
+  return gulp.src('lib/**').pipe(eslint({
+    'rules':{
+        'quotes': [1, 'single'],
+        'semi': [1, 'always']
```
mykmelez: I also recommend `comma-dangle: 'always-multiline'` and `quote-props: 'as-needed'` (both of which would entrain changes to this object literal).

lib/configure.js
```diff
@@ -321,7 +320,7 @@ module.exports = function() {
 
     function travisAwaitSyncing() {
       process.stdout.write('Waiting for Travis to finish syncing…\n');
-      return new Promise(function(resolve, reject) { setTimeout(resolve, 5000) })
+      return new Promise(function(resolve, reject) { setTimeout(resolve, 5000); })
```
mykmelez, line 323: I disagree with `semi: 'always'` for single-line single-statement function expressions like this one, which are akin to lambdas, as noted by [Arrow Functions - Shorter functions](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions#Shorter_functions). I find the semicolonless variant more readable. But there's no way to configure ESLint to accommodate this preference, and I'd much rather enforce semicolons `always` than `never`, so let it be!

marco-c, line 323: Indeed, in particular with the function calls that follow this one.

darkwing:
```
find . -name 'multiparty' -print
./node_modules/gulp-connect/node_modules/connect/node_modules/multiparty
```

Odd, as https://www.npmjs.com/package/gulp-connect hasn't been updated in a year.

darkwing:
Woot, failures issue fixed!

gulpfile.js
```diff
@@ -67,9 +68,20 @@ gulp.task('pre-test', function () {
...
-})
+});
+
+gulp.task('lint', function() {
+  return gulp.src('lib/**').pipe(eslint({
+    'rules':{
+        'quotes': [1, 'single'],
+        'semi': [1, 'always'],
+        'comma-dangle': [1, 'always-multiline'],
+        'quote-props': [1, 'as-needed']
```
mykmelez, line 84: Ah, right, we're linting `lib/**`, not `**`, so these quoted properties and the lack of a dangling comma here doesn't trigger the warnings.

mykmelez:
I will want to expand it over time, but this is a great start!

PR #141, opened 2015-11-05
animate spinner while awaiting async operations in configure

mykmelez:
The _configure_ command engages in a variety of asynchronous operations as it initiates requests to the GitHub and Travis API endpoints and awaits their responses. While it's in a waiting state, it should use an animated spinner to indicate as much to the user, so it's more obvious that everything is well, and the user remains apprised of the state of the process.

This work-in-progress adds a spinner to a single async operation, the deletion of an existing GitHub token, as a proof-of-concept for this enhancement. Next I'll add them to all such operations.

(Here I'm also experimenting with using a _WIP_ label rather than a prefix on the issue title to indicate the work-in-progress status of this pull request. Feedback on that is also welcome!)

### DIFF
--- a/lib/configure.js
+++ b/lib/configure.js
@@ -23,6 +23,7 @@ var childProcess = require('child_process');
 var ghslug = promisify(require('github-slug'));
 var promptly = require('promisified-promptly');
 var readYaml = require('read-yaml');
+var cli = require('cli');
 var travisEncrypt = promisify(require('travis-encrypt'));
 var writeYaml = require('write-yaml');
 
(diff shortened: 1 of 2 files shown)


marco-c:
> (Here I'm also experimenting with using a WIP label rather than a prefix on the issue title to indicate the work-in-progress status of this pull request. Feedback on that is also welcome!)

I was confused at first (when I read the word "work-in-progress" in the previous paragraph and I recalled the title didn't contain the string "WIP", I tried to see whether you changed the title from "WIP: XXX" to "XXX"), but I like it.

mykmelez:
Closing in favor of rolling this work into #149.

PR #147, opened 2015-11-05
Resolve #95 - Update Oghliner app content

### LINKED ISSUE #95: redesign Oghliner website
The Oghliner website at https://mozilla.github.io/oghliner/ was designed using the GitHub Pages generator, and it could use a redesign (both graphic design and information design) in order to better convey what the project is about and how to go about using it.

This used to be identical to #7, because we were using the website as the template for bootstrapping new apps. But now we've separated those two things, so the website is in app/, while the _bootstrap_ command uses the template in templates/. Thus redesigning the website is now a separate concern from redesigning the template.

darkwing:
This is just a start, will be building upon this.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -21,14 +21,25 @@ var gulp = require('gulp');
 var mocha = require('gulp-mocha');
 var istanbul = require('gulp-istanbul');
 var eslint = require('gulp-eslint');
+var template = require('gulp-template');
+var marked = require('marked');
 var argv = require('yargs').argv;
+
 var oghliner = require('./index.js');
 
 gulp.task('default', ['build', 'offline']);
 
-gulp.task('build', ['copy-files', 'build-tabzilla']);
+gulp.task('build', ['build-app', 'build-tabzilla']);
 
-gulp.task('copy-files', function(callback) {
+gulp.task('build-app', ['copy-files'], function() {
+  return gulp.src('dist/index.html')
+    .pipe(template({
+      content: marked(require('fs').readFileSync('README.md', 'utf8'))
+    }))
+    .pipe(gulp.dest('dist'));
+});
+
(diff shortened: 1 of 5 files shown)


darkwing:
This is ready for preliminary review.  I need to figure out a way to get rid of the buttons though.

marco-c:
I like the idea of reusing the README (idea that made me remember that we need to keep the README updated, filed #153).
I also like the result. I'd only add a bit more space between the sections (e.g. before "Using The Template").

gulpfile.js
```diff
@@ -21,14 +21,25 @@ var gulp = require('gulp');
...
 var oghliner = require('./index.js');
 
 gulp.task('default', ['build', 'offline']);
 
-gulp.task('build', ['copy-files', 'build-tabzilla']);
+gulp.task('build', ['build-app', 'build-tabzilla']);
+
+gulp.task('build-app', ['copy-files'], function() {
+  return gulp.src('dist/index.html')
+              .pipe(template({
```
marco-c: Nit: To keep the same style as the rest of the file:

```
  return gulp.src('dist/index.html')
    .pipe(template({
    ...
```

marco-c:
I also think the information could be reorganized a little in the README, to make it simpler to read.

darkwing:
Updates made to separate sections and update code style.

marco-c:
Last remaining issue:
screen

darkwing:
I tried `white-space: pre-line` but we were still getting overflow so the best solution is hiding overflow so code/commands look the same and we never see screen stretching.

marco-c:
Can we use `overflow: auto` so the scrollbars are only shown if needed?

darkwing:
/me facepalms

Of course, sorry.  No more early morning PRs from me

marco-c:
:smile:

PR #149, opened 2015-11-06
clarify output of configure command (fixes #148)

### LINKED ISSUE #148: clarify output of configure command
The _configure_ command's output should be clearer. This is part of #26. Here's a proposed redesign for the happiest path. Unfortunately, it isn't possible to set colors, but I would make the info section (which starts with a ℹ character) blue and the warning section (which starts with a ⚠ character) yellow and use some text formatting (color tone? bold weight?) to set apart the command example.

Each async operation message would also display a spinner to its right until it completes, at which point the text "done!" would replace it, and the checkmarks to the left of those messages would appear. We might color those checkmarks green, à la the checkmarks next to the test names in the test output.

```
Configuring Travis to auto-deploy [REPOSITORY SLUG] to GitHub Pages…

To check the status of your repository in Travis and authorize Travis to push
to it, I'll create GitHub personal access tokens, for which I need your GitHub
username and password (and two-factor authentication code, if appropriate).

ℹ For more information about GitHub personal access tokens, see:
    https://github.com/settings/tokens

Username: username
Password: 

You're using two-factor authentication with GitHub.
Please enter the code provided by your authentication software.

Auth Code: 123456

✓ Creating temporary GitHub token for getting Travis token… done!
✓ Getting Travis token… done!
✓ Deleting temporary GitHub token for getting Travis token… done!
✓ Creating permanent GitHub token for Travis to push to the repository… done!
✓ Checking the status of your repository in Travis… done!

Good news, your repository is already active in Travis!

✓ Encrypting permanent GitHub token… done!
✓ Writing configuration to .travis.yml file… done!

⚠ You didn't already have a .travis.yml file, so I created one for you.
  For more information about the file, see:
    http://docs.travis-ci.com/user/customizing-the-build/

You're ready to auto-deploy using Travis!  Just commit the changes
in .travis.yml and push the commit to the origin/master branch:

git commit -m"configure Travis to auto-deploy to GitHub Pages" .travis.yml
git push origin master

Then visit [TRAVIS BUILD URL FOR REPOSITORY] to see build status.
```

mykmelez:


mykmelez:
This work-in-progress also encompasses the spinner POC in #141, which I'll probably close in favor of landing the entire set of clarifications in this branch.

mykmelez:
@marco-c 1675349 resulted in an assertion failure in _syncs Travis with GitHub, sync was already in progress but finished before we checked and the repo is not found_. I'm not sure why, but I looked at the test, and it seems like the new behavior is what I would expect, so I updated the test. Can you take a look at it too and let me know what you think?

marco-c:
The error message should be `Sync already in progress. Try again later.` because that is what's thrown by `travis.users.sync.post`, I'm not sure why 1675349 changed that.

The code is:

```
         return travis.users.sync.post()
        .catch(function(err) {
          // Ignore the exception if Travis is already syncing. The status code
          // of the error is 409 and its message is 'Sync already in progress.
          // Try again later.', but instead of relying on that we simply perform
          // an explicit check.
          return travisIsSyncing().then(function(isSyncing) {
            if (!isSyncing) {
              // Check again if it's active on Travis (in case it finished syncing
              // right before we checked)
              return ensureActiveInTravis().catch(function() {
                // Throw the original error.
                throw err;
              });
            }
          });
        })
```

So the original error, the one thrown by travis.users.sync.post, should be thrown.

I think it would be fine if we didn't throw the original error but the error thrown by ensureActiveInTravis, but the fact that 1675349 didn't change this part of the code but changed its behavior makes me think that there's something else going on. Could you check what parts of this code are executed in the _syncs Travis with GitHub, sync was already in progress but finished before we checked and the repo is not found_ test?

We can then make configure throw the error thrown by `ensureActiveInTravis`, but I'd like to know what's going on at the moment.

mykmelez:
> Could you check what parts of this code are executed in the syncs Travis with GitHub, sync was already in progress but finished before we checked and the repo is not found test?

Yes, the difference was that the _travisIsSyncing_ check used to return `false`, but with my changes in 1675349 it was returning `true`.

> We can then make configure throw the error thrown by ensureActiveInTravis, but I'd like to know what's going on at the moment.

It turned out to be a leftover network mock from a previous test, which was apparently responding to the _travisIsSyncing_ request.

_nock_ lets you assert that all mocks have been used at the end of a test, but testConfigure hasn't been doing that, as it would have required some refactoring to expose the _scope_ object returned by the _nock_ functions.

nock version 2.17 makes it possible to check this for all scopes simultaneously by calling _nock.isDone_, however, so I updated the version of nock and added a check to _afterEach_ in e2f36e3. That identified the leftover mocks, which I've removed. And that let me revert the change to the test.

mykmelez:
Ok, this is ready for review. Here's a screenshot of a run with some fairly common code paths being exercised (repository already active in Travis, .travis.yml file already exists):

screen shot 2015-11-09 at 15 16 29

Note: per the design in #148, _configure_ asks for your auth code right after asking for your username/password, before it tells you that it's creating a token. To ensure that happens, I added a GET /authorizations call after the username/password prompt to trigger a `Must specify two-factor authentication OTP code.` error if the user is using 2FA. The result of the request doesn't matter, only that it throws an error when appropriate. It could be any request that requires authorization; that one just seems the simplest and most obvious.

package.json
```diff
@@ -31,6 +31,8 @@
     "templates/"
   ],
   "dependencies": {
+    "chalk": "^1.1.1",
```
marco-c, line 34: Can we use gutil.colors instead of chalk? So we don't need to depend on another package.

marco-c:
Nice layout!

lib/configure.js
```diff
@@ -66,6 +68,13 @@ travis.users.sync.post = promisify(travis.users.sync.post);
 // when creating GitHub tokens.
 var noteUrl = 'https://github.com/mozilla/oghliner';
 
+// The message currently displayed with a spinner, if any.  Most async
+// operations don't need to set this, but occasionally one displays a message
+// and then calls a function that displays another message before
+// the operation is complete.  In that case, the function needs to stop
+// the spinner for the first message before it displays the second message.
+var spinnerMessage;
```
marco-c: Could you create a function that does this automatically? Maybe by wrapping cli.spinner.

test/testConfigure.js
```diff
@@ -475,7 +492,17 @@ describe('Configure', function() {
...
+    })
+    .then(complete);
+  });
+
+  it('creates permanent GitHub token', function() {
+    nockBasicPostAuthFlow();
+    configure();
+    return enterUsernamePassword()
+    .then(function() {
+      return await('Creating permanent GitHub token for Travis to push to the repository… done!');
```
marco-c: Nit: maybe we could collapse some of these tests in a single one that checks all the output?

test/testConfigure.js
```diff
@@ -757,6 +826,9 @@ describe('Configure', function() {
   afterEach(function() {
     process.chdir(oldWd);
     temp.cleanupSync();
+    if (!nock.isDone()) {
```
marco-c, line 820: This is very useful.

marco-c:
> Ok, this is ready for review. Here's a screenshot of a run with some fairly common code paths being exercised (repository already active in Travis, .travis.yml file already exists):

Maybe we can reduce the amount of messages we print? The user probably doesn't care about it, unless there's something wrong and he needs to file an issue.

(P.S.: I mean the amount of messages printed while enabling the repo on Travis)

lib/configure.js
```diff
@@ -279,32 +303,75 @@ module.exports = function() {
...
-            return promisify(travis.hooks(hook.id).put)({ hook: { active: true } });
+
+            var message = 'Your repository isn\'t active in Travis yet; activating it…';
+            cli.spinner('  ' + message);
+            return promisify(travis.hooks(hook.id).put)({ hook: { active: true } })
+            .then(function(res) {
+              if (res.result) {
+                cli.spinner(chalk.green('✓ ') + message + ' done!', true);
+              } else {
+                cli.spinner(chalk.red('✓ ') + message + ' failed!', true);
```
marco-c: Maybe we can use a 'X' here instead of a 'V'?

package.json
marco-c, line 34: Or, let's use `chalk` everywhere (in a follow-up), since in most "clarifications" we're removing gutil.log.

lib/configure.js
```diff
@@ -456,8 +516,10 @@ module.exports = function() {
       'You\'re ready to auto-deploy using Travis!  Just commit the changes\n' +
       'in .travis.yml and push the commit to the origin/master branch:\n' +
       '\n' +
-      'git commit -m"configure Travis to auto-deploy to GitHub Pages" .travis.yml\n' +
-      'git push origin master\n' +
+      chalk.bold.gray(
```
marco-c: Gray is a little hard to read on my screen for some reason. It is too dark (darker than the screenshot you provided.

marco-c: What do you think of `chalk.bold.italic.white`? Too similar to the normal style?

marco-c: Or we could use one of the bg\* colors: https://raw.githubusercontent.com/chalk/ansi-styles/master/screenshot.png

marco-c: This is with `chalk.bold.italic.bgBlack`:
integrate

marco-c: This is with `chalk.bold.bgBlack`:
integrate2

marco-c: With `chalk.bold`:
integrate3

package.json
mykmelez, line 34: Filed as #167.

lib/configure.js
mykmelez: > Could you create a function that does this automatically? Maybe by wrapping cli.spinner.

I've been looking into adopting the one you created for _integrate_, but this case is trickier, because we want to abort the current spinner message, display a new one, then re-display the original one after the new one is done.

Which is not ideal output anyway, since it's a bit strange to leave the original spinner message hanging, without either a checkmark/done or x-mark/error indicator. So I've been thinking about a different design that displays only a single spinner message but also displays a note if the token creation required deletion of an existing token (like .travis.yml file creation displays a warning if it had to create the file), something like:

```
✓ Deleting temporary GitHub token for getting Travis token… done!
✓ Creating permanent GitHub token for Travis to push to the repository… done!

ℹ You had an existing token for this app, so we deleted and recreated it.
```

My first attempt at this is in e642a27, but it doesn't look quite right yet, as it results in:

```
✓ Deleting temporary GitHub token for getting Travis token… done!

ℹ You had an existing token for this app, so we deleted and recreated it.

✓ Creating permanent GitHub token for Travis to push to the repository… done!
```

mykmelez: 16ddb6b completes this work, making the messages appear in the correct order:

```
✓ Deleting temporary GitHub token for getting Travis token… done!
✓ Creating permanent GitHub token for Travis to push to the repository… done!

ℹ You had an existing token for this app, so we deleted and recreated it.
```

However, while implementing it, I realized that there are two other circumstances under which we'll want to pause/interrupt the spinner to display another message:

First, on an authentication failure. A username/password failure isn't likely to occur now that we check their values before displaying a message with a spinner. But an auth code failure can still occur when the current auth code expires, and that can happen during any operation that uses the GitHub API and requires authentication:

```
You're using two-factor authentication with GitHub.
Please enter the code provided by your authentication software.

Auth Code: 123456

  Creating temporary GitHub token for getting Travis token…
Your authentication code is incorrect or has expired.
Please try again.

Auth Code: 987654
```

Second, when we tell Travis to sync with GitHub and then display `Waiting for Travis to finish syncing…` messages every five seconds:

```
I didn't find your repository in Travis.  Syncing Travis with GitHub…
Waiting for Travis to finish syncing…
Waiting for Travis to finish syncing…
```

For the latter case, we can solve the problem by making `Syncing Travis with GitHub…` a spinner message and then not printing `Waiting for Travis to finish syncing…`. We only display that message every five seconds now as a kind of poor-man's spinner, so the user knows that activity is ongoing, and the program isn't frozen. We can replace it with a real spinner.

But we should still have a plan for the former case, so we do something sensible when we need to pause/interrupt a spinner temporarily to request a new auth code.

test/testConfigure.js
mykmelez: > Nit: maybe we could collapse some of these tests in a single one that checks all the output?

Yes, 44b6e66deda0863667883076226921ec3f8ac52a does this. I use Promise.all, which doesn't guarantee the order in which the promises get resolved, so I guess this test could produce a false positive if the messages appear out-of-order, but they're all there. Unsure how important that is.

mykmelez:
> (P.S.: I mean the amount of messages printed while enabling the repo on Travis)

Can you be more specific? Checking a repo's status is one line, and activating it is another single line:

```
✓ Checking the status of your repository in Travis… done!
✓ Your repository isn't active in Travis yet; activating it… done!
```

If your repo is already active, then you get an info message that takes up a few more, but it's still just a single statement:

```
✓ Checking the status of your repository in Travis… done!

Good news, your repository is already active in Travis!
```

Or perhaps you're talking about syncing Travis with GitHub? That can add lines:

```
✓ Checking the status of your repository in Travis… done!
I didn't find your repository in Travis.  Syncing Travis with GitHub…
Waiting for Travis to finish syncing…
Waiting for Travis to finish syncing…
✓ Checking the status of your repository in Travis… done!
```

But I'll get rid of the `Waiting for Travis to finish syncing…` lines per https://github.com/mozilla/oghliner/pull/149#discussion_r44480981.

lib/configure.js
mykmelez: > Maybe we can use a 'X' here instead of a 'V'?

Good idea, implemented in df62a56 (where I also change "failed!" to "error!" for consistency with _integrate_).

mykmelez:
> But I'll get rid of the Waiting for Travis to finish syncing… lines per #149 (comment).

Done in e73730b.

There's some strange behavior in this part of the flow with the test _syncs Travis with GitHub, sync was already in progress but finished before we checked and the repo is found_. On master, it produces the convoluted and confusing sequence:

```
Checking the status of your repository in Travis…
I didn't find your repository in Travis.  Syncing Travis with GitHub…
Checking the status of your repository in Travis…
Good news, your repository is already active in Travis!
Waiting for Travis to finish syncing…
Checking the status of your repository in Travis…
Good news, your repository is already active in Travis!
```

On this branch, before e73730b, it produced:

```
✓ Checking the status of your repository in Travis… done!
I didn't find your repository in Travis.  Syncing Travis with GitHub…
✓ Checking the status of your repository in Travis… done!

Good news, your repository is already active in Travis!

Waiting for Travis to finish syncing…
✓ Checking the status of your repository in Travis… done!

Good news, your repository is already active in Travis!
```

With e73730b, it produces:

```
✓ Checking the status of your repository in Travis… done!
✓ I didn't find your repository in Travis; syncing Travis with GitHub… done!
✓ Checking the status of your repository in Travis… done!

Good news, your repository is already active in Travis!

✓ I didn't find your repository in Travis; syncing Travis with GitHub… done!
✓ Checking the status of your repository in Travis… done!

Good news, your repository is already active in Travis!
```

So something is up with the code that ensures that the repository is active in Travis.

mykmelez:
The remaining work here is to pause/interrupt a spinner message while handling an authentication failure, per https://github.com/mozilla/oghliner/pull/149#discussion_r44480981, and to address the duplicate/confusing messages during _syncs Travis with GitHub, sync was already in progress but finished before we checked and the repo is found_, per https://github.com/mozilla/oghliner/pull/149#issuecomment-155613355.

(That second issue is an existing one, so I could file and fix it separately, but I'm touching code in the area, so I'll take a stab at it here first.)

marco-c:
> Can you be more specific? Checking a repo's status is one line, and activating it is another single line:

I meant that we could collapse all the text after "Auth Code:" and before "Encrypting permanent GitHub token" into a single "Activating repository on Travis...".

marco-c:
I've re-added the WIP label, this way I'll easily notice when you're done with the other changes.

mykmelez:
> pause/interrupt a spinner message while handling an authentication failure, per #149 (comment)

This is addressed by 93395fe and 5dddc91.

mykmelez:
While working on the last two changes, I noticed that the output for _prompts you to re-enter an incorrect username/password_ was strange:

```
To check the status of your repository in Travis and authorize Travis to push
to it, I'll create GitHub personal access tokens, for which I need your GitHub
username and password (and two-factor authentication code, if appropriate).

ℹ For more information about GitHub personal access tokens, see:
    https://github.com/settings/tokens

Username: username
Password: 

✓ Creating temporary GitHub token for getting Travis token… done!
✓ Getting Travis token… done!
✓ Deleting temporary GitHub token for getting Travis token… done!
  Creating permanent GitHub token for Travis to push to the repository…

The username and/or password you entered is incorrect; please re-enter them.

Username: username
Password: 
```

It turns out that I didn't update the test function when I added the credential check earlier on this branch. While updating it, I noticed that the credential check also doesn't re-check credentials after a failure. So I fixed that too. And added a "Checking credentials…" spinner to let the user know what the command is doing and make the formatting of credential failure messages more consistent.

The fixes are in 2f49daf, and now that test's output looks like:

```
To check the status of your repository in Travis and authorize Travis to push
to it, I'll create GitHub personal access tokens, for which I need your GitHub
username and password (and two-factor authentication code, if appropriate).

ℹ For more information about GitHub personal access tokens, see:
    https://github.com/settings/tokens

Username: username
Password: 

× Checking credentials… error!

The username and/or password you entered is incorrect; please re-enter them.

Username: username
Password: 

× Checking credentials… error!

The username and/or password you entered is incorrect; please re-enter them.

Username: username
Password: 

✓ Checking credentials… done!
```

mykmelez:
> I meant that we could collapse all the text after "Auth Code:" and before "Encrypting permanent GitHub token" into a single "Activating repository on Travis...".

I can imagine collapsing the three messages around getting a Travis token:

```
✓ Creating temporary GitHub token for getting Travis token… done!
✓ Getting Travis token… done!
✓ Deleting temporary GitHub token for getting Travis token… done!
```

Those could simply be:

```
✓ Getting Travis token… done!
```

Since the temporary GitHub token that we use to get the Travis token is ephemeral, so the user shouldn't see it elsewhere (modulo errors while getting the Travis token), and they don't have to know that we created and destroyed it.

But I wouldn't collapse the message about getting a permanent GitHub token, as I think it's important for developers to know that that token was created, since the token gives a third-party access to their repositories.

Unsure about the messages about activating the repository in Travis. It seems useful to let users follow along during the activation process, especially if we have to sync Travis with GitHub, as that can take a while. But even if not, since the Travis API endpoint seems slower than the GitHub one, so more fine-grained messages means we don't leave the developer hanging as long between messages. I could be convinced otherwise, however.

marco-c:
Thinking about it more, since configuration is supposed to happen just once (or just a few times anyway), there's no need to summarize too much.

So, I would collapse the messages around getting a Travis token, like you said, and I would collapse:

```
Creating permanent GitHub token for Travis to push to the repository...
You already have that token; deleting it... done!
Creating permanent GitHub token for Travis to push to the repository... done!
```

into:

```
Creating permanent GitHub token for Travis to push to the repository... done!
```

Instead of printing "You already have that token; deleting it", we could prompt the user if he wants to delete it (like the comment https://github.com/mozilla/oghliner/blob/12f6381e8085ea78aeb6912c2500ddec4dc2c91e/lib/configure.js#L161suggests). We can do this in a follow-up though.

mykmelez:
> There's some strange behavior in this part of the flow with the test _syncs Travis with GitHub, sync was already in progress but finished before we checked and the repo is found_.

The duplicate messages from _ensureActiveInTravis_ (`Checking the status of your repository in Travis…` and `Good news, your repository is already active in Travis!`) happen because that code path results in duplicate calls to that function (and _travisAwaitSyncing_). Essentially, the path does this:
- ensureActiveInTravis -> throws `repository not found`
- travis.users.sync.post -> throws `Sync already in progress. Try again later.`
- travisIsSyncing -> returns `false` (no longer syncing)
- ensureActiveInTravis -> returns `undefined` (success)
- travisIsSyncing -> returns `false` (not syncing)
- ensureActiveInTravis -> returns `undefined` (success)

The version on this branch also calls _stopSpinner_ before each call to _ensureActiveInTravis_, in order to stop the spinner for `I didn\'t find your repository in Travis; syncing Travis with GitHub…`, which is why that message gets duplicated.

I'll think on how to fix this.

mykmelez:
> I'll think on how to fix this.

4bf2462 reorganizes the logic for handling `repository not found` errors so _ensureActiveInTravis_ (and _travisAwaitSyncing_) is only re-invoked once after Travis finishes syncing. The behavior should otherwise remain the same.

It also adds a test to ensure that _configure_ fails with `repository not found` when a sync is successful, but the repository still isn't found.

mykmelez:
> Instead of printing "You already have that token; deleting it", we could prompt the user if he wants to delete it (like the comment https://github.com/mozilla/oghliner/blob/12f6381e8085ea78aeb6912c2500ddec4dc2c91e/lib/configure.js#L161suggests). We can do this in a follow-up though.

Yes, we might want to prompt at that point, unsure. In the meantime, I'm a little wary of removing the message, because I'm not sure if there are ramifications to deleting the token.

I think there aren't, since Oghliner should be the only tool that requests that token, and it never reveals it to the user, so there's little chance that the user will re-use it. (I've only ever re-used it by hacking Oghliner to log it, and only to debug Oghliner.) But I'm not yet confident enough to say that we should elide that detail of the process.

At the very least, it could be useful for the developer to know that they already had a token, which means they already configured the repo once. And also that the old token is no longer valid, so they can no longer use it to auto-deploy, even if they discard the changes that _configure_ makes to .travis.yml.

Anyway, let's tackle this in a followup.

mykmelez:
@marco-c: I removed the _WIP_ label, as I _think_ I've addressed all the various concerns (yours and mine) here. But please double-check! This one is a complicated set of changes.

marco-c:
The only thing missing is collapsing the three messages around getting a Travis token, but let's do that in a follow-up.

PR #159, opened 2015-11-09
Support choosing which remote you want to publish to

### LINKED ISSUE #54: make it easier to configure auto-deploy for *upstream* repo
The _configure_ command assumes that you want to auto-deploy merges to the _master_ branch on the _origin_ remote.

But for projects with multiple contributors, it's common for a user to fork the original repo and clone their fork, such that the _origin_ remote points to their fork, and then create an _upstream_ remote for the original repo, as described in GitHub's [Fork A Repo](https://help.github.com/articles/fork-a-repo/) documentation.

And it's actually more interesting to configure auto-deploy for the original, upstream repo, since that's the one in which people will more often merge changes via pull requests and want those merges to trigger auto-deploy.

But it's hard to do that right now. You have to either temporarily rename your remotes so the original remote is named _origin_ (which actually sounds like a better name for it, although it is unconventional) or clone the original remote and configure that clone.

We should make it easier to configure auto-deploy for the _upstream_ repo.

For example, we could prompt the user to choose the remote for which to configure auto-deploy, selecting the _upstream_ remote by default if one is defined.

(A related problem is that you can't configure a repo to auto-deploy both the original repo and a fork, since the encrypted GitHub auth token in .travis.yml is repo-specific, and there can be only one auth token in the Travis build environment variables. But that's a different issue.)

marco-c:
Fixes #54.


marco-c:
I'd wait #119 to land first.

marco-c:
Closing in favor of #165.

PR #165, opened 2015-11-10
Support choosing which remote you want to publish to

marco-c:


lib/configure.js
```diff
@@ -439,15 +439,31 @@ module.exports = function() {
...
+    var remotes = childProcess.execSync('git remote').toString().split('\n').slice(0, -1);
+    if (remotes.indexOf('upstream') !== -1) {
+      defaultRemote = 'upstream';
+    }
+
+    var promise;
+    if (remotes.length === 1) {
+      promise = Promise.resolve(remotes[0]);
+    } else {
+      promise = promptly.prompt('Repository remote [default: ' + defaultRemote + ']: ', { default: defaultRemote });
```
mykmelez: Nit: Repository remote -> Remote repository

test/testConfigure.js
```diff
@@ -754,6 +754,44 @@ describe('Configure', function() {
...
+
+    return enterUsernamePassword()
+    .then(function() {
+      return await('Repository remote [default: upstream]:');
+    })
+    .then(function() {
+      emit('origin\n');
+    })
+    .then(complete);
+  });
```
mykmelez, line 857: It'd also be useful to check that the default value is _origin_ if there is more than one remote and none of them are called _upstream_.

test/testLiveAsTemplate.js
```diff
@@ -124,6 +124,10 @@ describe('CLI interface, oghliner as a template', function() {
     .then(function() {
       assert.doesNotThrow(fse.statSync.bind(fse, 'index.html'));
       assert.doesNotThrow(fse.statSync.bind(fse, 'offline-worker.js'));
+    })
+    .catch(function(err) {
+      console.log('Error: ' + err);
+      throw err;
```
mykmelez: Is this (and the one below) just for testing, or is it permanent? And if the latter, doesn't it cause duplicate messages to get written to the terminal, one to stdout, the other to stderr?

mykmelez: I did a quick test, throwing an error in the previous link of the chain, and it didn't display the message twice, although it did display the word "Error" twice, i.e. `Error: Error: text of error message`. Presumably this is because `new Error('text of error message').toString()` produces `Error: text of error message`, so `console.log('Error: ' + err)` duplicates the `Error:` prefix.

marco-c: Yes, for some reason mocha doesn't print to the console the rejection error in this case... So I've explicitly added logging, otherwise we don't know why the test fails when it fails.

I've tried writing a simpler test:

```
describe('a', function() {
  it('test', function() {
    return Promise.reject(new Error('message'));
  });
});
```

And it logs:

```
  a
    1) test


  0 passing (12ms)
  1 failing

  1) a test:
     Error: message
      at Context.<anonymous> (test/testRejection.js:3:27)
```

So I'll look into it more.

marco-c:
Filed #170 to use the default gh-pages module when (and if) https://github.com/tschaub/gh-pages/pull/24 gets merged.

test/testConfigure.js
marco-c, line 857: Done in 67beaa9.

test/testLiveAsTemplate.js
marco-c: I've found the problem, opened #174 to fix it and reverted the changes in this branch.

marco-c: Basically, the getTokenId promise was never resolved (because it kept trying with higher and higher page numbers), so the mocha test was never ending (the promise in _after_ was never resolved).

lib/configure.js
```diff
@@ -71,13 +71,6 @@ module.exports = function() {
   // without having to pass them down the chain.
   var slug, user, repo, username, password, otpCode, token, tempToken, tempTokenId;
 
-  // XXX Will users sometimes want to configure Travis to deploy changes to a
-  // non-"origin" repository?  For example, origin is mykmelez/test-app, but I
-  // want to configure Travis to deploy changes to mozilla/test-app, which has
-  // a different remote name (like upstream).  If so, then we'll need to prompt
-  // the user to choose the remote for which to configure deployment. For now,
-  // though, we assume they are configuring the origin remote.
-
```
mykmelez, line 111: I just remembered the reason why this comment is all the way up here: the first thing this code does is set _slug_ to the slug for the _origin_ remote (using the _github-slug_ module). And we use that in the initial message (`Configuring Travis to auto-deploy mozilla/oghliner to GitHub Pages…`) as well the token notes (`Oghliner token for mozilla/oghliner`, `temporary Oghliner token to get Travis token for mozilla/oghliner`) and, most importantly, when encrypting the GitHub token for use by Travis (which requires you to specify the slug of the repository for which the token can be used).

(We also use it to set a default value for the _username_ field. We might want to add some smarts to guess the username better when there are both _origin_ and _upstream_ remotes, since the username is probably the one from the _origin_ remote, even if the user is auto-deploying the _upstream_ remote.)

So we should determine the repository for which we're configuring auto-deploy up here, before we have to use it. I wouldn't prompt before displaying the initial message, however, since that would be too abrupt, and it wouldn't give the user a chance to learn what the command will do first.

Thus, I'd remove the slug from the initial message and display something like this (basing this on the new language in #149):

```
Configuring Travis to auto-deploy to GitHub Pages.

Your repository has multiple remotes; please specify the one to auto-deploy.

Remote [default: upstream]: 

Ok, I'll configure Travis to auto-deploy the [REMOTE] remote ([SLUG FOR REMOTE]).
```

Or, if there's only one remote:

```
Configuring Travis to auto-deploy app to GitHub Pages.

Your repository has a single remote, [REMOTE] ([SLUG FOR REMOTE]).
That's what I'll configure Travis to auto-deploy.
```

(Wordsmithing welcome!)

lib/configure.js
```diff
@@ -439,15 +432,31 @@ module.exports = function() {
...
+      promise = Promise.resolve(remotes[0]);
+    } else {
+      promise = promptly.prompt('Remote repository [default: ' + defaultRemote + ']: ', { default: defaultRemote });
     }
 
-    writeYaml.sync('.travis.yml', travisYml);
+    return promise.then(function(remote) {
+      command = 'echo "travis_fold:end:after_success" && ' +
+                '[ "${TRAVIS_PULL_REQUEST}" = "false" ] && [ "${TRAVIS_BRANCH}" = "master" ] && ' +
+                'echo "Deploying…" && gulp deploy --remote ' + remote;
```
mykmelez: Although the remote to deploy can be different on a user's local machine, it should always be _origin_ on Travis, given how Travis clones the repo.

marco-c: Right, fixed in a87e5cd.

marco-c:
I think this is basically done, I only want to clean up the code that gets the slug (maybe refactoring it into a separate npm module).

lib/bootstrap.js
```diff
@@ -98,16 +83,16 @@ function getDefaultTemplateConfig(dir) {
...
-    if (url) {
-      config.repository = url;
-      // Try to fill in the project named based on the repo url.
-      if (url.substr(-4, 4) === '.git') {
-        config.name = url.substring(url.lastIndexOf('/') + 1, url.length - 4);
-      }
+  return gitRemoteUrl(dir, 'origin')
+  .then(function(url) {
+    config.repository = url;
+    // Try to fill in the project named based on the repo url.
```
mykmelez, line 89: Nit: named -> name

PR #168, opened 2015-11-11
Clarify bootstrap command

marco-c:
Fixes #160.

PR #172, opened 2015-11-11
Remove gulp-util dependency

marco-c:
After #168 and #171, gulp-util will be unused.


mykmelez:
Waiting on this one until #168 is merged.

PR #176, opened 2015-11-11
Clarify `deploy` command

### LINKED ISSUE #163: clarify output of deploy command
The _deploy_ command's output could be clearer. Here's an example of the current design:

```
[16:21:44] Deploying "ignore cache created when using oghliner CLI"
[16:21:44] Cloning git@github.com:mykmelez/eggtimer.git into .gh-pages-cache
[16:21:44] Cleaning
[16:21:44] Fetching origin
[16:21:45] Checking out origin/gh-pages
[16:21:45] Removing files
[16:21:45] Copying files
[16:21:45] Adding all
[16:21:45] Committing
[16:21:45] Pushing
```

As with #162, this is basically fine, modulo unnecessary timestamps. Perhaps we'd want to add an ellipsis (…) to the end of each step, then prepend a green checkmark and append the word "done!" after the step is complete, like for _configure_ in #148.

I'd also update the header line to look like:

```
Deploying "ignore cache created when using oghliner CLI" to GitHub Pages…
```

And if the commit message is multi-line, I'd ensure that only the first line appears in that header line.

As with #162, this command runs frequently and often is automated as part of a build process, so we should avoid extraneous header/footer information.

marco-c:
Fixes #163.

Depends on #172.


marco-c:
Example output:
deploy

test/testDeploy.js
```diff
@@ -254,4 +255,47 @@ describe('Deploy', function() {
...
+
+        if (output.indexOf('Deploying "' + expected + '" to GitHub Pages…') !== -1) {
+          process.stdout.write = write;
+          resolve();
+        }
+      };
+    });
+  }
+
+  it('should prasd', function() {
```
mykmelez: It should _prasd_?

marco-c: Oops, I forgot to actually write that :smile:

marco-c: Done in 8503a7f.

PR #181, opened 2015-11-12
Update glob version range

marco-c:
Instead of replacing `^5.0.15` by `^6.0.1`, I've used the range `^5.0.15 || ^6.0.1`.
This way an oghliner user forced to use 5.x for some reason could still install oghliner.

Is my understanding correct?


mykmelez:
Hmm, yes, I think this is correct.

PR #184, opened 2015-11-13
rename project (fixes #23) — DON'T MERGE!

### LINKED ISSUE #23: consider renaming module to something descriptive
We should consider renaming the module to something descriptive, like _offline-gh-pages_, per [these best practices for naming Node modules](https://github.com/mattdesl/module-best-practices#naming-conventions).

On the other hand, command-line utilities benefit from being short and sweet, and Oghliner exposes a command-line interface, in addition to its module interface. So perhaps it makes sense to keep the existing, compact name.

mykmelez:
This is what it'd look like to rename the project to _offline-github-pages_ while calling the command-line tool _oghp_. This is a complete fix for #23 (the only test failure is due to _offline-github-pages_ not yet existing in npm), but I've labeled it WIP (and marked it `DON'T MERGE!`) because I'm still not sure that we should do it.

I'll decide by tomorrow, but at the moment, after looking at this, I'm leaning against changing the name, since it's easier to remember the name of the command-line tool if it's the same name as the package (and project), and offline-github-pages is long and cumbersome as a name for a command-line tool.


mykmelez:
Per https://github.com/mozilla/oghliner/issues/23#issuecomment-156575287, we're not going to do this after all.

PR #186, opened 2015-11-13
Use native promises in sw-precache - DON'T MERGE

marco-c:


marco-c:
Still uninformative errors.

PR #175, opened 2015-11-11
Clarify the output of the `offline` command

marco-c:
Fixes #162.

Depends on #172.

Issue #162, opened 2015-11-10
clarify output of offline command


mykmelez:
The _offline_ command's output could be clearer. Here's the current output:

```
[16:15:17] Caching static resource "dist/CNAME" (13 B)
[16:15:17] Caching static resource "dist/css/index.css" (378 B)
[16:15:17] Caching static resource "dist/favicon/android-chrome-144x144.png" (2.65 kB)
[16:15:17] Caching static resource "dist/favicon/android-chrome-192x192.png" (3.65 kB)
[16:15:17] Caching static resource "dist/favicon/android-chrome-36x36.png" (692 B)
[16:15:17] Caching static resource "dist/favicon/android-chrome-48x48.png" (839 B)
[16:15:17] Caching static resource "dist/favicon/android-chrome-72x72.png" (1.23 kB)
[16:15:17] Caching static resource "dist/favicon/android-chrome-96x96.png" (1.67 kB)
[16:15:17] Caching static resource "dist/favicon/apple-touch-icon-114x114.png" (2.06 kB)
[16:15:17] Caching static resource "dist/favicon/apple-touch-icon-120x120.png" (2.15 kB)
[16:15:17] Caching static resource "dist/favicon/apple-touch-icon-144x144.png" (2.73 kB)
[16:15:17] Caching static resource "dist/favicon/apple-touch-icon-152x152.png" (2.9 kB)
[16:15:17] Caching static resource "dist/favicon/apple-touch-icon-180x180.png" (3.47 kB)
[16:15:17] Caching static resource "dist/favicon/apple-touch-icon-57x57.png" (1 kB)
[16:15:17] Caching static resource "dist/favicon/apple-touch-icon-60x60.png" (1.05 kB)
[16:15:17] Caching static resource "dist/favicon/apple-touch-icon-72x72.png" (1.25 kB)
[16:15:17] Caching static resource "dist/favicon/apple-touch-icon-76x76.png" (1.29 kB)
[16:15:17] Caching static resource "dist/favicon/apple-touch-icon-precomposed.png" (4.87 kB)
[16:15:17] Caching static resource "dist/favicon/apple-touch-icon.png" (3.47 kB)
[16:15:17] Caching static resource "dist/favicon/browserconfig.xml" (383 B)
[16:15:17] Caching static resource "dist/favicon/favicon-16x16.png" (522 B)
[16:15:17] Caching static resource "dist/favicon/favicon-194x194.png" (4.87 kB)
[...]

Issue #24, opened 2015-10-06
make service worker ignore query parameters when matching URL to resource in cache


mykmelez:
Oghliner doesn't set sw-precache's _ignoreUrlParametersMatching_ option, and that option's default value of `[/^utm_/]` means that sw-precache takes most query parameters into account when deciding whether or not to serve a resource from its cache.

That's a problem for apps like presentations using [Shower](https://github.com/shower/shower), which set query parameters like _full_ and _lang_ in client code to store state in a way that is bookmarkable. If a user loads an app page from a URL with a query parameter that wasn't present when the app was initially offlined, then the service worker will not serve the page from the cache, so the load will fail if the user is offline.

To solve that problem, _offline_ should set sw-precache's _ignoreUrlParametersMatching_ option to `[/./]` so it ignores all query parameters when matching URLs to fetch against URLs in its cache.

(Being able to do this for all apps automagically is one of the benefits of focusing on the GitHub Pages use case. Since GitHub Pages sites are always static, and query parameters never affect the result of a GET request for a GitHub Pages resource, we can always do this when offlining apps specifically for deployment on GitHub Pages.)

marco-c:
As you said, this is specific to static sites. I agree we should only support deploying to GitHub Pages for now, but at the same time I think we shouldn't force anything on people who want to use oghliner only for offlining. So I'd make this configurable, maybe enabled by default, and we could show a warning when someone uses `deploy` without this option.

PR #188, opened 2015-11-13
Public domain sw

delapuente:
This PR removes the dependency with sw-precache at the same time it provides a basic servie worker.

### DIFF
--- a/lib/offline.js
+++ b/lib/offline.js
@@ -20,7 +20,6 @@
 var promisify = require('promisify-node');
 
 var fs = require('fs');
-var conflict = require('gulp-conflict');
 var path = require('path');
 var chalk = require('chalk');
 var glob = require('glob');
@@ -51,7 +50,7 @@ module.exports = function(config) {
     var importScripts = config.importScripts || [];
 
     // Remove the existing service worker, if any, so the worker doesn't include
-    // it in the list of files to cache.
+    // itself in the list of files to cache.
     try {
       fs.unlinkSync(path.join(rootDir, 'offline-worker.js'));
     } catch (ex) {
@@ -62,14 +61,16 @@ module.exports = function(config) {
       }
     }
 
-    ghslug('./').catch(function() {
+    ghslug('./')
+    .catch(function() {
       // Use the name from package.json if there's an error while fetching the GitHub slug.
       try {
         return JSON.parse(fs.readFileSync('package.json')).name;
       } catch (ex) {
         return '';
       }
-    }).then(function(cacheId) {
+    })
+    .then(function(cacheId) {
       checkImports(importScripts, rootDir);
 
       var absoluteGlobs = fileGlobs.map(function (v) {
@@ -80,7 +81,7 @@ module.exports = function(config) {
 
       var replacements = {
         cacheId: cacheId,
-        ignoreUrlParametersMatching: config.ignoreUrlParametersMatching || [/./], //TODO: Not implemented in template
+        ignoreUrlParametersMatching: config.ignoreUrlParametersMatching || [/./],
         cacheVersion: getHashOfHashes(pluckHashes(filesAndHashes)),
         resources: filesAndHashes,
         importScripts: importScripts,
@@ -88,14 +89,14 @@ module.exports = function(config) {
 
       var stream = gulp.src([__dirname + '/../templates/app/offline-worker.js'])
         .pipe(template(replacements))
-        .pipe(conflict(rootDir))
         .pipe(gulp.dest(rootDir));
 
       return new Promise(function (resolve, reject) {
         stream.on('finish', function () { resolve(); });
         stream.on('error', function (e) { reject(e); });
       });
-    }).then(resolve, reject);
+    })
+    .then(resolve, reject);
 
     function flatGlobs(fileGlobs) {
       return Object.keys(fileGlobs.reduce(function (matches, fileGlob) {
@@ -110,10 +111,11 @@ module.exports = function(config) {
     function getFilesAndHashes(files, rootDir, sizeWhiteList) {
       var totalSize = 0;
       var filesAndHashes = files.map(function (filepath) {
-        totalSize += countFileSize(filepath, sizeWhiteList);
+        var size = countFileSize(filepath, sizeWhiteList);
+        totalSize += size;
         var data = fs.readFileSync(filepath);
         var hash = getHash(data);
-        logOk(filepath);
+        logOk(filepath, size);
         return {
           path: filepath.replace(rootDir, function (match, offset) {
             //XXX: The root must be the worker's directory
@@ -123,15 +125,14 @@ module.exports = function(config) {
         };
       });
 
-      console.log('Total precache size is about ' + prettyBytes(totalSize) +
-                  ' for ' + files.length + ' resources.');
+      console.log('Total cache size is ' + prettyBytes(totalSize) +
+                  ' for ' + files.length + ' files.\n');
 
       return filesAndHashes;
     }
 
     function checkImports(imports, rootDir) {
       imports.forEach(function (filepath) {
-        //TODO: I think we should warn, not fail
         assertExists(path.join(rootDir, filepath));
       });
     }
@@ -163,8 +164,8 @@ module.exports = function(config) {
       return stat.size;
     }
 
-    function logOk(filepath) {
-      console.log(chalk.bold.green('✓ ') + 'Caching ' + filepath);
+    function logOk(filepath, size) {
+      console.log(chalk.bold.green('✓ ') + 'Caching ' + filepath + ' (' + prettyBytes(size) + ')');
     }
 
     function getHashOfHashes(hashArray) {
--- a/lib/offline.js
+++ b/lib/offline.js
@@ -70,7 +70,7 @@ module.exports = function(config) {
         return '';
       }
     }).then(function(cacheId) {
-      var importsAndHashes = getImportsAndHashes(importScripts, rootDir);
+      checkImports(importScripts, rootDir);
 
       var absoluteGlobs = fileGlobs.map(function (v) {
         return path.join(rootDir, v);
@@ -81,9 +81,9 @@ module.exports = function(config) {
       var replacements = {
         cacheId: cacheId,
         ignoreUrlParametersMatching: config.ignoreUrlParametersMatching || [/./], //TODO: Not implemented in template
-        cacheVersion: getVersion(filesAndHashes),
+        cacheVersion: getHashOfHashes(pluckHashes(filesAndHashes)),
         resources: filesAndHashes,
-        importScripts: importsAndHashes,
+        importScripts: importScripts,
       };
 
       var stream = gulp.src([__dirname + '/../templates/app/offline-worker.js'])
@@ -123,28 +123,21 @@ module.exports = function(config) {
         };
       });
 
-      console.log('Total precache size is about ' + prettyBytes(totalSize));
+      console.log('Total precache size is about ' + prettyBytes(totalSize) +
+                  ' for ' + files.length + ' resources.');
+
       return filesAndHashes;
     }
 
-    function getImportsAndHashes(imports, rootDir) {
-      return imports
-        .map(function (filepath) {
+    function checkImports(imports, rootDir) {
+      imports.forEach(function (filepath) {
+        //TODO: I think we should warn, not fail
+        assertExists(path.join(rootDir, filepath));
+      });
+    }
 
-          filepath = path.join(rootDir, filepath);
-          //TODO: I think we should warn, not fail
-          assertExists(filepath);
-
-          var data = fs.readFileSync(filepath);
-          var hash = getHash(data);
-          return {
-            path: filepath.replace(rootDir, function (match, offset) {
-              //TODO: Are we sure about this replacement is not relative to the worker directory
-              return offset === 0 ? '' : match;
-            }),
-            hash: hash,
-          };
-        });
+    function pluckHashes(entries) {
+      return entries.map(function (entry) { return entry.hash; });
     }
 
     function assertExists(filepath) {
@@ -160,12 +153,6 @@ module.exports = function(config) {
         console.log(chalk.red.bold(filepath + ' is not a file.'));
         throw new Error(filepath + ' is not a file.');
       }
-    }
-
-    function getVersion(filesAndHashes) {
-      return getHashOfHashes(filesAndHashes.map(function(entry) {
-        return entry.hash;
-      }));
     }
 
     function countFileSize(filepath, whiteList) {
--- a/lib/offline.js
+++ b/lib/offline.js
@@ -28,9 +28,12 @@ var template = require('gulp-template');
 var crypto = require('crypto');
 var gulp = require('gulp');
 var ghslug = promisify(require('github-slug'));
+var prettyBytes = require('pretty-bytes');
 
 module.exports = function(config) {
   return new Promise(function(resolve, reject) {
+    'use strict';
+
     var rootDir = config.rootDir || './';
 
     console.log('Offlining ' + chalk.bold(rootDir) + ' to ' + chalk.bold(path.join(rootDir, 'offline-worker.js')) + '…\n');
@@ -41,11 +44,11 @@ module.exports = function(config) {
     // directory for a custom domain (f.e. https://example.com/) or a subdirectory
     // of a GitHub Pages domain (f.e. https://mykmelez.github.io/example/).
     if (rootDir.lastIndexOf('/') !== rootDir.length -1) {
-      rootDir = rootDir + "/";
+      rootDir = rootDir + '/';
     }
 
     var fileGlobs = config.fileGlobs || ['**/*'];
-    var importGlobs = config.importScripts || [];
+    var importScripts = config.importScripts || [];
 
     // Remove the existing service worker, if any, so the worker doesn't include
     // it in the list of files to cache.
@@ -67,21 +70,20 @@ module.exports = function(config) {
         return '';
       }
     }).then(function(cacheId) {
+      var importsAndHashes = getImportsAndHashes(importScripts, rootDir);
+
       var absoluteGlobs = fileGlobs.map(function (v) {
         return path.join(rootDir, v);
       });
       var files = flatGlobs(absoluteGlobs);
       var filesAndHashes = getFilesAndHashes(files, rootDir, absoluteGlobs);
 
-      var imports = flatGlobs(importGlobs);
-      var importsAndHashes = getImportsAndHashes(imports,   
-
       var replacements = {
-        cacheId: cacheId, //TODO: Not implemented in template
+        cacheId: cacheId,
         ignoreUrlParametersMatching: config.ignoreUrlParametersMatching || [/./], //TODO: Not implemented in template
         cacheVersion: getVersion(filesAndHashes),
         resources: filesAndHashes,
-        importScripts: importsAndHashes
+        importScripts: importsAndHashes,
       };
 
       var stream = gulp.src([__dirname + '/../templates/app/offline-worker.js'])
@@ -89,9 +91,11 @@ module.exports = function(config) {
         .pipe(conflict(rootDir))
         .pipe(gulp.dest(rootDir));
 
-      stream.on('finish', function () { resolve(); });
-      stream.on('error', function (e) { reject(e); });
-    });
+      return new Promise(function (resolve, reject) {
+        stream.on('finish', function () { resolve(); });
+        stream.on('error', function (e) { reject(e); });
+      });
+    }).then(resolve, reject);
 
     function flatGlobs(fileGlobs) {
       return Object.keys(fileGlobs.reduce(function (matches, fileGlob) {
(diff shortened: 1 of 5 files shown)


delapuente:
@mykmelez or @marco-c , do you want to take a look? 
I missed sw-precache is able to take a template parameter to generate. I'm preparing another PR not removing the sw-precache but with the custom template. What do you think?

mykmelez:
> @mykmelez or @marco-c , do you want to take a look?

You bet! Looking at this now…

> I missed sw-precache is able to take a template parameter to generate. I'm preparing another PR not removing the sw-precache but with the custom template. What do you think?

That's a good idea, since we're about to release v1, and using sw-precache with a custom template is a smaller set of changes, so it should be less risky. (We'll probably still remove sw-precache eventually, so I would keep this PR open and rebase it once we land that one.)

lib/offline.js
```diff
@@ -42,8 +48,9 @@ module.exports = function(config) {
     }
 
     var fileGlobs = config.fileGlobs || ['**/*'];
+    var importScripts = config.importScripts || [];
 
-    // Remove the existing service worker, if any, so sw-precache doesn't include
+    // Remove the existing service worker, if any, so the worker doesn't include
     // it in the list of files to cache.
```
mykmelez: Nit: with this change, I'd also modify the sentence to read, "so the worker doesn't include _itself_ in the list of files to cache."

lib/offline.js
```diff
@@ -63,52 +70,111 @@ module.exports = function(config) {
...
       });
+      var files = flatGlobs(absoluteGlobs);
+      var filesAndHashes = getFilesAndHashes(files, rootDir, absoluteGlobs);
 
-      staticFileGlobs.forEach(function(globPattern) {
-        glob.sync(globPattern.replace(path.sep, '/')).forEach(function(file) {
-          var stat = fs.statSync(file);
+      var replacements = {
+        cacheId: cacheId,
+        ignoreUrlParametersMatching: config.ignoreUrlParametersMatching || [/./], //TODO: Not implemented in template
```
mykmelez: Rather than labeling this line TODO, I'd remove it and file a followup issue to reimplement it in the new template. It isn't important at the moment, since GitHub Pages only supports static content. Although it might be useful in the future for offlining an app and then hosting it with another hosting provider (per #24).

lib/offline.js
```diff
@@ -63,52 +70,111 @@ module.exports = function(config) {
...
-          }
-
-          console.log(message);
-        },
-        importScripts: importScripts,
-        cacheId: cacheId,
-        ignoreUrlParametersMatching: config.ignoreUrlParametersMatching || [/./],
-        maximumFileSizeToCacheInBytes: Infinity,
+      console.log('Total precache size is about ' + prettyBytes(totalSize) +
+                  ' for ' + files.length + ' resources.');
```
mykmelez: While we're here, let's update this message to make it a bit clearer. I suggest:

Total cache size is [TOTAL SIZE] for [NUMBER OF FILES] files.

Reasoning:
- precache -> cache (I think I understand what sw-precache docs mean when they say that the service worker "precaches" resources, but I'm not sure it's useful to make the distinction, especially without defining it further. And the Oghliner docs say that the service worker "caches" the files. Plus the actual API the Service Worker uses is the Cache API, and this script says it is "Caching" each file. So "cache" is the better word to use here.)
- ~~about~~ (Sure, a "pretty" number isn't accurate to an exact number of bytes, but it's still accurate enough, so "about" feels like unnecessary complexity.)
- resources -> files (Oghliner docs talk about assets and files, while sw-precache docs talk about resources and files. In both cases, the things being cached are files, and assets/resources is a more generic euphemism. Let's cut to the chase.)

lib/offline.js
```diff
@@ -63,52 +70,111 @@ module.exports = function(config) {
...
+
+      return filesAndHashes;
+    }
+
+    function checkImports(imports, rootDir) {
+      imports.forEach(function (filepath) {
+        //TODO: I think we should warn, not fail
```
mykmelez: Can you file this as a followup issue?

templates/app/offline-worker.js
```diff
@@ -0,0 +1,126 @@
+/* Any copyright is dedicated to the Public Domain.
```
mykmelez, line 1: I wonder if this should live at the top level of the directory tree, since it's a template, but unlike the other files in templates/, this one doesn't get copied to a developer's app directory when they bootstrap an app. Eh, it probably doesn't matter.

lib/offline.js
```diff
@@ -63,52 +70,111 @@ module.exports = function(config) {
...
+      };
 
-          if (stat.isFile() && stat.size > 2 * 1024 * 1024 && staticFileGlobs.indexOf(file) === -1) {
-            console.log(chalk.yellow.bold(file + ' is bigger than 2 MiB. Are you sure you want to cache it? To suppress this warning, explicitly include the file in the fileGlobs list.'));
-          }
-        });
+      var stream = gulp.src([__dirname + '/../templates/app/offline-worker.js'])
+        .pipe(template(replacements))
+        .pipe(conflict(rootDir))
+        .pipe(gulp.dest(rootDir));
```
mykmelez, line 91: Doing this with Gulp has the unfortunate side-effect of displaying a complex and confusing error message at the end of the process:

```
[13:39:15] [conflict] Creating offline-worker.js
```

We shouldn't display this message, per the design proposal in #162 and its implementation in #175. (If we did decide that it made sense to tell the user we were creating that file, then it would look very different, something like `✓ Creating PATH/TO/offline-worker.js`.) Can you suppress that output?

templates/app/offline-worker.js
```diff
@@ -0,0 +1,126 @@
...
+    CACHE_PREFIX: 'offline-cache:<%= cacheId %>:' + (self.registration ? self.registration.scope : '') + ':',
+
+    // This is the unique name for the cache controlled by this version of the worker.
+    get CACHE_NAME() {
+      return this.CACHE_PREFIX + '<%= cacheVersion %>';
+    },
+
+    // This is a list of resources that will be cached.
+    RESOURCES: [
+      '/',
```
mykmelez: Is this to cache the default page?

mykmelez: (If so, nice! Would be worth a comment, just to make it extra-obvious.)

lib/offline.js
```diff
@@ -55,7 +62,7 @@ module.exports = function(config) {
       }
     }
 
-    resolve(ghslug('./').catch(function() {
+    ghslug('./').catch(function() {
```
mykmelez: Nit: while you're changing this line, can you move the .catch dereferencing/invocation to the next line? That'd make it more consistent with the style used elsewhere in this codebase, which is based on the style in the [Q library](https://github.com/kriskowal/q) documentation.

``` js
    ghslug('./')
    .catch(function() {
```

templates/app/offline-worker.js
```diff
@@ -0,0 +1,126 @@
...
+<% }); %>
+    ],
+
+    // Adds the resources to the cache controlled by this worker.
+    cacheResources: function () {
+      var _this = this;
+      var now = Date.now();
+      var baseUrl = self.location;
+      return _this.openCache()
+        .then(function (cache) {
```
mykmelez: Nit: here too (and below) I would use the style we use elsewhere in this codebase, which is very similar to the style you use here, except that the `.then(…` dereferencing/invocation isn't indented:

``` js
      return _this.openCache()
      .then(function (cache) {
```

templates/app/offline-worker.js
```diff
@@ -0,0 +1,126 @@
...
+        .then(function (cacheNames) {
+          return Promise.all(cacheNames.map(deleteIfNotCurrent));
+        });
+
+      function deleteIfNotCurrent(cacheName) {
+        if (cacheName.indexOf(_this.CACHE_PREFIX) !== 0 || cacheName === _this.CACHE_NAME) {
+          return Promise.resolve();
+        }
+        return self.caches.delete(cacheName);
+      }
```
mykmelez: Nit: instead of defining a __this_ alias for _this_, use `Function.bind(this)` to bind these functions to the correct _this_ object.

templates/app/offline-worker.js
```diff
@@ -0,0 +1,126 @@
...
+              if (response.ok) {
+                return cache.put(originalRequest, response);
+              }
+              console.error('Error fetching ' + url + ', status was ' + response.status);
+            });
+          }));
+        });
+    },
+
+    // Remove the offline caches non controlled by this worker.
```
mykmelez: Nit: non -> not

lib/offline.js
```diff
@@ -63,52 +70,111 @@ module.exports = function(config) {
...
+    function countFileSize(filepath, whiteList) {
+      var stat = fs.statSync(filepath);
+      if (stat.isFile() && stat.size > 2 * 1024 * 1024 && whiteList.indexOf(filepath) === -1) {
+        console.log(chalk.yellow.bold(filepath + ' is bigger than 2 MiB. Are you sure you want to cache it? To suppress this warning, explicitly include the file in the fileGlobs list.'));
+      }
+      return stat.size;
+    }
+
+    function logOk(filepath) {
+      console.log(chalk.bold.green('✓ ') + 'Caching ' + filepath);
```
mykmelez: This doesn't put quotes around the file path, as the current implementation does, but I like this version better, since the quotes don't add much, as it's already easy to distinguish the path from the word "Caching" that appears before it.

Also missing here is the size of the file in parentheses after the path. I have mixed feelings about that change, but I'm ok with it, as it isn't clear how useful that information is.

lib/offline.js
delapuente: Totally agree.

templates/app/offline-worker.js
delapuente, line 1: As you want. If we don't like this way, we can open a follow up later.

templates/app/offline-worker.js
delapuente: Sure.

lib/offline.js
delapuente, line 91: I simply removed the warning as the information about the sw being created is displayed before the caching messages.

delapuente:
All done, @mykmelez do you mind to take a second look? I also changed [how the worker prepares the new cache](https://github.com/mozilla/oghliner/pull/188/files#diff-15733a51ce60a106273263fd073ab695R62) to be populated: before adding content, it deletes it in case the last installation failed and we would have a partially populated cache.

lib/offline.js
```diff
@@ -55,60 +61,121 @@ module.exports = function(config) {
...
+        .pipe(template(replacements))
+        .pipe(gulp.dest(rootDir));
 
-          if (stat.isFile() && stat.size > 2 * 1024 * 1024 && staticFileGlobs.indexOf(file) === -1) {
-            console.log(chalk.yellow.bold(file + ' is bigger than 2 MiB. Are you sure you want to cache it? To suppress this warning, explicitly include the file in the fileGlobs list.'));
-          }
-        });
+      return new Promise(function (resolve, reject) {
+        stream.on('finish', function () { resolve(); });
+        stream.on('error', function (e) { reject(e); });
```
marco-c: Nit: no need to define another function, simply do `stream.on('finish', resolve);`

lib/offline.js
```diff
@@ -55,60 +61,121 @@ module.exports = function(config) {
...
+
+      var absoluteGlobs = fileGlobs.map(function (v) {
         return path.join(rootDir, v);
       });
+      var files = flatGlobs(absoluteGlobs);
+      var filesAndHashes = getFilesAndHashes(files, rootDir, absoluteGlobs);
+
+      var replacements = {
+        cacheId: cacheId,
+        ignoreUrlParametersMatching: config.ignoreUrlParametersMatching || [/./],
```
mykmelez: Nit: I would remove this entirely, so no one is misled into thinking it has an effect.

lib/offline.js
```diff
@@ -55,60 +61,121 @@ module.exports = function(config) {
...
+      var totalSize = 0;
+      var filesAndHashes = files.map(function (filepath) {
+        var size = countFileSize(filepath, sizeWhiteList);
+        totalSize += size;
+        var data = fs.readFileSync(filepath);
+        var hash = getHash(data);
+        logOk(filepath, size);
+        return {
+          path: filepath.replace(rootDir, function (match, offset) {
+            //XXX: The root must be the worker's directory
```
mykmelez: Typically I use _XXX_ comments to identify issues that need resolution, i.e. it's the equivalent of _TODO_. But it isn't clear what issue is being identified by this comment. Can you provide more detail here? Or, even better, file a followup issue on it!

lib/offline.js
```diff
@@ -55,60 +61,121 @@ module.exports = function(config) {
...
+    function countFileSize(filepath, whiteList) {
+      var stat = fs.statSync(filepath);
+      if (stat.isFile() && stat.size > 2 * 1024 * 1024 && whiteList.indexOf(filepath) === -1) {
+        console.log(chalk.yellow.bold(filepath + ' is bigger than 2 MiB. Are you sure you want to cache it? To suppress this warning, explicitly include the file in the fileGlobs list.'));
+      }
+      return stat.size;
+    }
+
+    function logOk(filepath, size) {
+      console.log(chalk.bold.green('✓ ') + 'Caching ' + filepath + ' (' + prettyBytes(size) + ')');
```
mykmelez: Was this originally being called from multiple call sites? I'm just wondering why it's factored out into a function when it's a one-liner that is only being called from one site.

templates/app/offline-worker.js
```diff
@@ -0,0 +1,129 @@
...
+      }
+    }));
+  });
+
+  // Retrieves the request following oghliner strategy.
+  self.addEventListener('fetch', function (event) {
+    if (event.request.method === 'GET') {
+      event.respondWith(oghliner.get(event.request));
+    }
+    else {
```
mykmelez: Nit: cuddle the _else_ keyword, i.e.:

```
} else {
```

(I personally prefer elses uncuddled, the way you've written this one; but else-cuddling is conventional in this codebase, as it is in a bunch of other Mozilla codebases, and I'd rather we keep it consistent.)

templates/app/offline-worker.js
```diff
@@ -0,0 +1,129 @@
...
+      event.respondWith(self.fetch(event.request));
+    }
+  });
+
+  var oghliner = self.oghliner = {
```
mykmelez, line 35: I don't see _self.oghliner_ being used anywhere. Is this to support imported scripts having access to the _oghliner_ object?

templates/app/offline-worker.js
```diff
@@ -0,0 +1,129 @@
...
+      .then(function (cacheNames) {
+        return Promise.all(cacheNames.map(deleteIfNotCurrent.bind(this)));
+      }.bind(this));
+
+      function deleteIfNotCurrent(cacheName) {
+        if (cacheName.indexOf(this.CACHE_PREFIX) !== 0 || cacheName === this.CACHE_NAME) {
+          return Promise.resolve();
+        }
+        return self.caches.delete(cacheName);
+      }
```
mykmelez: Note: if you bind the _deleteIfNotCurrent_ function itself, then you don't have to also bind the anonymous function in the promise chain:

``` js
      var deleteIfNotCurrent = function (cacheName) {
        if (cacheName.indexOf(this.CACHE_PREFIX) !== 0 || cacheName === this.CACHE_NAME) {
          return Promise.resolve();
        }
        return self.caches.delete(cacheName);
      }.bind(this);

      return self.caches.keys()
      .then(function (cacheNames) {
        return Promise.all(cacheNames.map(deleteIfNotCurrent));
      });
```

mykmelez: (We should eventually replace these bindings with arrow functions, which should be available everywhere service workers are available.)

mykmelez:
@delapuente Is this still a work-in-progress? It seems pretty ready, but the issue summary still has a _WIP_ tag.

lib/offline.js
```diff
@@ -55,60 +61,121 @@ module.exports = function(config) {
...
+        totalSize += size;
+        var data = fs.readFileSync(filepath);
+        var hash = getHash(data);
+        logOk(filepath, size);
+        return {
+          path: filepath.replace(rootDir, function (match, offset) {
+            //XXX: The root must be the worker's directory
+            return offset === 0 ? './' : match;
+          }),
+          hash: hash,
```
marco-c, line 123: Nit: you could avoid defining these two variables (`data` and `hash`) and directly do `hash: getHash(fs.readFileSync(filepath));', to make the function cleaner.

templates/app/offline-worker.js
```diff
@@ -0,0 +1,129 @@
...
+%>importScripts('<%- filepath %>');
+<% }); %>
+(function (self) {
+  'use strict';
+
+  // On install, cache resources and skip waiting so the worker won't
+  // wait for clients to be closed before becoming active.
+  self.addEventListener('install', function (event) {
+    event.waitUntil(oghliner.cacheResources().then(function () {
+      if (typeof self.skipWaiting === 'function') {
```
marco-c: Are there browsers that shipped without `skipWaiting`? If there aren't, then we can clean up this function by removing this.

templates/app/offline-worker.js
```diff
@@ -0,0 +1,129 @@
...
+        return self.skipWaiting();
+      }
+    }));
+  });
+
+  // On activation, delete old caches and start controlling the clients
+  // without waiting for them to reload.
+  self.addEventListener('activate', function (event) {
+    event.waitUntil(oghliner.clearOtherCaches().then(function () {
+      if (self.clients && typeof self.clients.claim === "function") {
```
marco-c: Same as before.

marco-c:
I did a second pass today, it looks good to me. I wish we had tests that actually executed the service worker code, but the changes look safe enough to land to me.

delapuente:
My onboarding session finished. I'm going to remove the WIP tag, fix all the nits and I will open a follow up for testing the service worker, how do you want to test the service worker? We can use [karma-sw-mocha](https://github.com/delapuente/karma-sw-mocha) as a testing environment if you want. It's being used in [ServiceWorkerWare](https://github.com/gaia-components/serviceworkerware/blob/master/testing/karma-sw.conf.js).

lib/offline.js
delapuente: I use XXX to highlight something important to remember, I can simply remove the XXX.

lib/offline.js
delapuente: It was but no more. Removing...

templates/app/offline-worker.js
delapuente, line 35: Yes, to expose the object and to allow better testing.

lib/offline.js
mykmelez: You could replace it with "NB" (from the Italian "nota bene", i.e. "note well") or simply "Note", but I'd still be interested in more details about what this comment is noting, as I don't quite understand the point being made.

templates/app/offline-worker.js
delapuente: You're right, I prefer to put dependencies below the code using them but here we save a binding. Let's do this way.

lib/offline.js
mykmelez: Erm, where by Italian I mean Latin of course! :wink:

delapuente:
@mykmelez for some reason, this is not working in the latest Nightly but it's doing perfectly in Chrome. I'm going to review the worker because I don't know what's happening.

lib/offline.js
marco-c: It is also a correct Italian phrase :smile:

marco-c:
> We can use karma-sw-mocha as a testing environment if you want. It's being used in ServiceWorkerWare.

Sounds good. I guess you can set it up pretty easily since you wrote it :smile:

delapuente:
Well, what does not work is controlled updating. The service worker got updated and installed but there is no `updatefound` event.

After some research, it turns out the sw-precache version is not working either. I think it could be related with [bug 1189659](https://bugzilla.mozilla.org/show_bug.cgi?id=1189659). I have some test cases (very simple workers) where I got it to run in Nightly but I fail making the oghliner version to work. I will be talking to bkelly or baku to solve this mistery.

Anyway, I think the new service worker is ready enough. What do you think @marco-c @mykmelez ?

lib/offline.js
```diff
@@ -55,60 +61,116 @@ module.exports = function(config) {
...
+      var totalSize = 0;
+      var filesAndHashes = files.map(function (filepath) {
+        var size = countFileSize(filepath, sizeWhiteList);
+        totalSize += size;
+        var data = fs.readFileSync(filepath);
+        var hash = getHash(data);
+        console.log(chalk.bold.green('✓ ') + 'Caching ' + filepath + ' (' + prettyBytes(size) + ')');
+        return {
+          path: filepath.replace(rootDir, function (match, offset) {
+            // The root must be the worker's directory
```
mykmelez, line 120: I'd still like to understand what this comment is suggesting, but I'm not going to block landing the branch on this.

mykmelez:
> Anyway, I think the new service worker is ready enough. What do you think @marco-c @mykmelez ?

I agree, I'm going to merge it, and then we can tackle any issues in followups. Thanks for all the iterations!

PR #191, opened 2015-11-14
New template

delapuente:
This version only adds a new template for sw-precahe (avoiding the cache per resource approach of the original template). The main problem here is how to keep the two cache versions until activation has been done as sw-precache does not identify the new worker as it does `cacheVersion` in pr #188.

The proposal use a temporal _new_ cache to populate it during installation. During activation, the contents of this _new_ cache are moved to the current. Personally, I don't like this solution as if for some reason, the activation process gets interrupted, the contents can be lost or be only partially updated. This can not occur in pr #188.

Another problem is that we need to keep the hashes as part of the actual code (not commentaries as in pr #188) because we need to retain the fingerprint of the content being cache in order to produce a byte-level change when updating the resources. In the pr #188 this is addressed by including a hash of all content hashes.

### DIFF
--- a/lib/bootstrap.js
+++ b/lib/bootstrap.js
@@ -118,7 +118,9 @@ module.exports = function(config) {
       console.log('\nCreating files…');
 
       return new Promise(function(resolve, reject) {
-        var stream = gulp.src([__dirname + '/../templates/**'])
+        var contents = __dirname + '/../templates/**';
+        var workerTemplate = __dirname + '/../templates/app/offline-worker.tmpl';
+        var stream = gulp.src([contents, '!' + workerTemplate])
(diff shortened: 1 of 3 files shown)


delapuente:
@mykmelez or @marco-c here you have the template version of the new worker? What do you think?

To avoid the problem described before, we should need to write some entries to indexed db to keep track of the process.

mykmelez:
Hmm, this is certainly a smaller set of code changes, but I'm inclined to merge #188, once it's ready, as it avoids the issues you noted, and it feels safe enough.

mykmelez:
We've landed #188 instead of this, so I'll close this.

PR #197, opened 2015-11-14
update readme to include info about all possible commands and options

### LINKED ISSUE #153: Update info in the README
The README doesn't contain all the info about all the possible oghliner commands, we should update it.

mykmelez:
Fixes #153.

### DIFF
--- a/README.md
+++ b/README.md
@@ -175,12 +175,12 @@ oghliner offline dist/
 To use it via the module, call `oghliner.offline`, passing it an *options* object with a *rootDir* property specifying the directory containing the files to offline:
 
 ```js
-oghliner.offline({
+var promise = oghliner.offline({
   rootDir: 'dist/',
 });
 ```
 
-If left unspecified, the default value of *rootDir* is `./`, i.e. the current directory.
+The call returns a Promise that resolves once Oghliner finishes generating the service worker. If left unspecified, the default value of *rootDir* is `./`, i.e. the current directory.
 
 Note: *rootDir* should be the *target* directory containing the output of your build process, not the *source* directory containing the original files. For example, if your source files are in *app/*, and your build process outputs into *dist/*, then you should specify the *dist/* directory.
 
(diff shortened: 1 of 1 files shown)


mykmelez:
This is now ready for review. You'll probably be better off reviewing the [new rendered version of the README](https://github.com/mykmelez/oghliner/blob/update-readme/README.md) rather than the patch itself, since I rewrote most of the file

marco-c:
I know you haven't touched this part, but it needs to be updated as well:

``` js
 var oghliner = require('oghliner');

 gulp.task('offline', function(callback) {
   oghliner.offline({
     rootDir: 'dist/',
     fileGlobs: [
       '**/*.html',
       'js/**/*.js',
     ],
   }, callback);
 });

 gulp.task('deploy', function(callback) {
   oghliner.deploy({
     rootDir: 'dist/',
   }, callback);
 });
```

The `offline` and `deploy` no longer have a callback as the second argument, so we should remove it from the example.
Gulp supports promises, so you can simply modify it to:

``` js
 var oghliner = require('oghliner');

 gulp.task('offline', function() {
   return oghliner.offline({
     rootDir: 'dist/',
     fileGlobs: [
       '**/*.html',
       'js/**/*.js',
     ],
   });
 });

 gulp.task('deploy', function() {
   return oghliner.deploy({
     rootDir: 'dist/',
   });
 });
```

README.md
```diff
@@ -1,56 +1,316 @@
...
+
+Then you can offline and deploy the app using the offline and deploy commands.
+
+ℹ For more information about offlining and deployment, see:
+    https://mozilla.github.io/oghliner/
+```
+
+Then add a &lt;script> tag referencing that script to your app's HTML page(s)/template(s). Oghliner will suggest one, but it doesn't necessarily know your app's directory structure, so make sure it contains the correct path to the file!
+
+## Offline
```
marco-c, line 165: I'd specify the `offline` and `deploy` return type (promise), it could be useful for developers.

marco-c:
Looks great!

mykmelez:
> I know you haven't touched this part, but it needs to be updated as well:

Ah, indeed, updated in 846fc43.

README.md
mykmelez, line 165: > I'd specify the offline and deploy return type (promise), it could be useful for developers.

Good idea, documented in 82cc1af.

PR #206, opened 2015-11-17
sync template's Oghliner dependency w/current version of Oghliner

mykmelez:
We should keep these in sync.


marco-c:
Filed #210 to do this automatically.

PR #207, opened 2015-11-17
make 'gulp build' and 'gulp offline' explicit build steps

mykmelez:
I'm not a fan of the way we obscure the build steps behind the _default_ task run by Gulp when you invoke it without arguments. If there were many steps, then I'd understand, but there are only two: _build_ and _offline_. We should run these two steps explicitly, and document them accordingly, so it's clear to developers how to build and offline a bootstrapped app.

This is based on #197, since it edits the README reference to gulp, and the README is being rewritten by that pull.


marco-c:
Looks good to me, I'll wait the landing of #197 before merging this.

PR #214, opened 2015-11-19
Update promisify-node dependency

marco-c:


mykmelez:
After upgrading _glob_ to 6.0.1, I see many warnings like this one when installing npm packages:

> npm WARN unmet dependency /usr/local/lib/node_modules/oghliner/node_modules/cli requires glob@'^5.0.10' but will load
> npm WARN unmet dependency /usr/local/lib/node_modules/oghliner/node_modules/glob,
> npm WARN unmet dependency which is version 6.0.1

I wonder if aggressively updating promisify-node will have a similar effect. And also if it'll cause any regressions, given that promisify-node's docs say its stable version is 0.1.5, and its build badge shows an error.

What's the benefit to updating this dependency?

marco-c:
> What's the benefit to updating this dependency?

It's mainly to tell that `oghliner` is compatible with `promisify-node` 0.3.0 and avoid issues when our users install `oghliner` alongside `promisify-node` 0.3.0.
I wish there was a way to tell `npm` which version we prefer, but avoid warnings when a user installs `oghliner` with a supported-but-not-preferred version of a package we depend on.

PR #221, opened 2015-11-21
clean up output (make it more consistent, reduce redundancy)

mykmelez:
This does a few things to make the output of the various commands more consistent. (Along the way, I do a bit of code cleanup too.) Some of the lines were changed multiple times while I was working on the branch, so it may be easier to review each commit individually.


mykmelez:
> This should be process.stderr.write.

Right, fixed in 2a9b4670d7f315c15202e3ca056da723da7612f8.

PR #224, opened 2015-11-25
Check for Updated version of SW on registration

vigneshshanmugam:


marco-c:
Thank you for the PR!
Why would you like to force checking for updates on page load?

vigneshshanmugam:
This makes sure the SW always caches the new files. 

Check here - https://slightlyoff.github.io/ServiceWorker/spec/service_worker/index.html#service-worker-registration-update-method

marco-c:
> This makes sure the SW always caches the new files. 

Yes, but the browser checks if there are updates periodically (every 24h) anyway. Calling `update` forces the checking to happen straight away.
I'm not sure it should be the default behavior, maybe we can add an option to specify if you want to always force checking or not.

@mykmelez @delapuente what do you think?

brendandahl:
The browser should also be checking at the end of every navigation event (on page load too), so I'm not sure it good to call this always. There's an issue with Firefox with this https://bugzilla.mozilla.org/show_bug.cgi?id=1226443

templates/app/scripts/offline-manager.js
```diff
@@ -27,6 +27,10 @@ function updateFound() {
 if ('serviceWorker' in navigator) {
   navigator.serviceWorker.register('offline-worker.js').then(function(registration) {
     console.log('offline worker registered');
+    // Check if there's an updated version of service-worker
```
delapuente, line 30: Please, remove the comment and the check (as all the browsers supporting SW have `update()` implemented) and put the `update()` invocation after registering the event.

delapuente:
It is not a bad idea as probably the developer expect from oghliner to update their files as soon as the assets change. @mykmelez a second opinion, please?

mykmelez:
> Yes, but the browser checks if there are updates periodically (every 24h) anyway. Calling update forces the checking to happen straight away.
> I'm not sure it should be the default behavior, maybe we can add an option to specify if you want to always force checking or not.

Checking for an update on load seems reasonable, for apps hosted on GitHub Pages, which I expect to be predominantly single-page apps (or apps with a small number of page loads), since apps hosted on GitHub Pages can't run server-side logic (because GitHub Pages only hosts static assets).

> The browser should also be checking at the end of every navigation event (on page load too), so I'm not sure it good to call this always.

Hmm, indeed, offline-manager.js calls the [register(scriptURL, options) method](https://slightlyoff.github.io/ServiceWorker/spec/service_worker/index.html#navigator-service-worker-register) on load, which runs the [Register algorithm](https://slightlyoff.github.io/ServiceWorker/spec/service_worker/index.html#register-algorithm), which runs the [Update algorithm](https://slightlyoff.github.io/ServiceWorker/spec/service_worker/index.html#update-algorithm), which is the same algorithm that the [update() method](https://slightlyoff.github.io/ServiceWorker/spec/service_worker/index.html#service-worker-registration-update) runs.

So it seems like this change would merely duplicate the existing update check that the browser itself does when offline-manager.js registers the service worker on load. Is there something I'm missing?

> There's an issue with Firefox with this https://bugzilla.mozilla.org/show_bug.cgi?id=1226443

It looks like there's a fix in hand for that issue. Even if there wasn't, however, I'd still take this fix (perhaps working around the Firefox issue) if it's otherwise worth taking.

delapuente:
True. Register runs Update algorithm so it's not necessary to call `update()` again. @vigneshshanmugam better file a bug if you find a problem updating the SW.

mykmelez:
Before we move on here, I want to make sure I understand the original motivation for suggesting this change.

@vigneshshanmugam: based on my reading of the spec, browsers should automatically check for an update each time a page within the service worker's scope is loaded, so it shouldn't be necessary to manually check for an update on load. Are you seeing different behavior, or is there another reason why it would make sense to call _update_ explicitly?

Note: @brendandahl filed [bug 1226443](https://bugzilla.mozilla.org/show_bug.cgi?id=1226443) on what appeared to be an intermittent failure to update the service worker but turned out to be an intermittent failure to call _onupdatefound_ after finding an update. If you don't see the browser update a service worker, and you're using that event to determine when an update is found, then you might be running into that problem.

vigneshshanmugam:
@mykmelez @delapuente Thanks guys, Will file a bug if its not getting updated. I always force it by this way in the register function to make sure the update is done on each registration. 

I followed the implementation of sw-precache which does the same check that I was suggesting. Not sure which one is correct though.

mykmelez:
> @mykmelez @delapuente Thanks guys, Will file a bug if its not getting updated. I always force it by this way in the register function to make sure the update is done on each registration.

Ah, good to know!

> I followed the implementation of sw-precache which does the same check that I was suggesting. Not sure which one is correct though.

Indeed, it looks like that was initially added in January of this year, a long time ago:

https://github.com/GoogleChrome/sw-precache/commit/1cf38966395f886c4f27a181b0aadac200ae9f24

I wonder if that was due to a bug in Chrome at that time.

vigneshshanmugam:
Hmmm might be. @slightlyoff @jakearchibald Any inputs here?

slightlyoff:
I know the API doesn't really communicate this clearly, but I think about the SW update/registration flow having 2 modes: "easy mode" and "roll your own" mode.

Easy mode is the default. In this mode, you generally put all of your App Shell asset caching logic inside the install step and rev the Service Worker whenever this changes. Because the SW changes when you update the asset list, the next navigation that checks for an update to the SW file will trigger installation of the new version. This works well because the activation of that version only takes place (by default) when you next visit the site. This activation strategy lets you prune your old caches and do database migrations (etc) when there's no other tab potentially fighting over those resources.

But Service Workers also let you implement roughly any/all of this logic yourself. `clients.claim()` and `skipWaiting()`, combined with fetch-time logic for populating caches (or `onmessage` handlers tickled from a document) allow you to do cache population totally outside the install phase, update and manage caches independently, and implement whatever policy you fancy for when to take over existing clients. These "roll your own" mode features are there to give you flexibility when the default policies don't match your app's needs.

The specific question about what oghliner should do seems, to me, to be related to the question of "is waiting until you visit the site again before you see updated content soon enough?" If the answer is "yes", than I think you just need to make sure that the SW file is changed in some way (e.g., a version number in a comment or variable value) and that you clean up old caches when you activate.

If oghliner wants to do something more custom, than I think calling `update()` might be useful but I'm not sure I understand the needs deeply enough to comment. Hope that helps.

delapuente:
@slightlyoff, correct me if I understood it wrong but `update()` is there to allow forcing installation and activation stages **without** actually change the SW?

slightlyoff:
`update` will not force those stages to run if the SW script itself had not
changed and the cache headers allow it to be used (I.e. max 24hr cache
expiry).

PR #226, opened 2015-11-28
Test some rootDir corner cases

marco-c:


mykmelez:
Nit: I'd probably name the test to clarify that _offline_ will create _rootDir_ if it doesn't already exist.

PR #230, opened 2015-11-30
Don't tell the user that we're creating a temporary GitHub token for getting Travis token

marco-c:


mykmelez:
If we remove the _Creating_ message, then we also need to remove the _Deleting_ message.

marco-c:
> If we remove the Creating message, then we also need to remove the Deleting message.

 Right, fixed in 3b4b4bb.

mykmelez:
Hmm, strange, "completes basic flow" in testConfigure.js should fail with this change, but it doesn't. That test needs updating, but we should also figure out why it passes unexpectedly. The latter can be a followup.

The example in the README also needs updating.

marco-c:
f7bd86a fixes the problem with the test.
4123061 removes the now unexpected messages from the test.
4114fe8 updates the README.

PR #234, opened 2015-12-04
Adding in directoryIndexes support for replacing filenames with dir c…

jonathanKingston:
…aching

This change is to add support for directories to be supported instead of the file that they are stored at in the file structure.

Currently in Chrome this causes an issue as it seems Firefox makes the assumption that the index.html resolves to the directory.
This can be seen on https://www.cyber-ami.com/how-we-help/ where chrome caches: https://www.cyber-ami.com/how-we-help/index.html (This was even the case when we were 301 redirecting the html file to the directory).

Either way it seems better form to cache the directory and have a setting to control it.

Let me know if there are any issues or queries thank you.

### DIFF
--- a/README.md
+++ b/README.md
@@ -190,6 +190,7 @@ The *offline* command takes the following options:
 
 - `--file-globs glob,…` or `fileGlobs: ['glob', …]` - a comma-separated list of globs identifying the files to offline (default: `**/*`). The globs are matched inside *rootDir*.
 - `--import-scripts script,…` or `importScripts: ['script', …]` - a comma-separated list of additional scripts to evaluate in the service worker (no default value). This is useful, for example, when you want to use the [Push API](https://developer.mozilla.org/en-US/docs/Web/API/Push_API).
+- `--directory-indexes index.html,…` or `directoryIndexes: ['index.html', …]` - a comma-separated list of file names that will be replaced to be served as a directory. (default: `index.html`) This fixes links to `'path/name/'` that don't have the file: `'path/name/index.html'`
 
 ## Deploy
 
--- a/lib/offline.js
+++ b/lib/offline.js
@@ -29,6 +29,7 @@ var gulp = require('gulp');
 var ghslug = promisify(require('github-slug'));
 var prettyBytes = require('pretty-bytes');
 var gzipSize = require('gzip-size');
+var escapeStringRegexp = require('escape-string-regexp');
 
 module.exports = function(config) {
   return new Promise(function(resolve, reject) {
@@ -51,6 +52,8 @@ module.exports = function(config) {
 
     var fileGlobs = config.fileGlobs || ['**/*'];
     var importScripts = config.importScripts || [];
+    var directoryIndexes = config.directoryIndexes || ['index.html'];
+    var directoryIndexMatch = new RegExp('/(' + directoryIndexes.map(escapeStringRegexp).join('|') + ')$');
 
     // Remove the existing service worker, if any, so the worker doesn't include
     // itself in the list of files to cache.
@@ -110,6 +113,15 @@ module.exports = function(config) {
       }, {}));
     }
 
+    function getPathName(filepath, rootDir) {
+       filepath = filepath.replace(directoryIndexMatch, '/');
+
+       return filepath.replace(rootDir, function (match, offset) {
+         // The root must be the worker's directory
+         return offset === 0 ? './' : match;
+       });
+    }
+
     function getFilesAndHashes(files, rootDir, sizeWhiteList) {
       var totalSize = 0;
       var totalGzipSize = 0;
@@ -119,19 +131,30 @@ module.exports = function(config) {
         var data = fs.readFileSync(filepath);
         totalGzipSize += gzipSize.sync(data);
         var hash = getHash(data);
-        process.stdout.write(chalk.green.bold('✓ ') + 'Caching ' + filepath + ' (' + prettyBytes(size) + ')\n');
+        var path = getPathName(filepath, rootDir);
+        var rootMatch = new RegExp('^' + rootDir);
+        var pathAtString = '';
+        if (filepath.replace(rootMatch, './') !== path) {
+          pathAtString = ' at path ' + path;
+        }
+
+        process.stdout.write(chalk.green.bold('✓ ') + 'Caching ' + filepath + pathAtString + ' (' + prettyBytes(size) + ')\n');
         return {
-          path: filepath.replace(rootDir, function (match, offset) {
-            // The root must be the worker's directory
-            return offset === 0 ? './' : match;
-          }),
+          path: path,
           hash: hash,
         };
       });
 
       process.stdout.write('Total cache size is ' + prettyBytes(totalSize) + ' (' + prettyBytes(totalGzipSize) + ' if served with gzip) for ' + files.length + ' files.\n');
 
-      return filesAndHashes;
+      filesAndHashes.unshift({
+        path: './'
+      }); // cache always the current root to make the default page available
+
+      // Remove duplicates from file list
+      return filesAndHashes.filter(function(fileName, index) {
+        return filesAndHashes.indexOf(fileName) === index;
+      });
     }
 
     function checkImports(imports, rootDir) {
--- a/test/testOffline.js
+++ b/test/testOffline.js
@@ -257,6 +257,31 @@ describe('Offline', function() {
     });
   });
 
+  it('should cache directoryIndexes as dirs', function() {
+    var rootDir = temp.mkdirSync('oghliner');
+    var dir = path.join(rootDir, 'dist');
+    fs.mkdirSync(dir);
+    var subDir = path.join(dir, 'subdir');
+    fs.mkdirSync(subDir);
+
+    fs.writeFileSync(path.join(subDir, 'test_file_1.js'), 'test_file_1');
+    fs.writeFileSync(path.join(subDir, 'test_file_2.js'), 'test_file_2');
+    fs.writeFileSync(path.join(subDir, 'test_file_3.js'), 'test_file_3');
+
+    process.chdir(rootDir);
+
+    return offline({
+      rootDir: dir,
+      directoryIndexes: ['test_file_1.js']
+    }).then(function() {
+      var content = fs.readFileSync(path.join(dir, 'offline-worker.js'), 'utf8');
+      assert.notEqual(content.indexOf('\'./\''), -1, 'Must have a directory in the output');
+      assert.equal(content.indexOf('test_file_1.js'), -1, 'Must have replaced the file name');
+      assert.notEqual(content.indexOf('test_file_2.js'), -1);
+      assert.notEqual(content.indexOf('test_file_3.js'), -1);
+    });
+  });
+
   it('should cache large files', function() {
     var rootDir = temp.mkdirSync('oghliner');
     var dir = path.join(rootDir, 'dist');
--- a/cli.js
+++ b/cli.js
@@ -94,11 +94,13 @@ program
   .description('offline the files in the directory by generating offline-worker.js script')
   .option('--file-globs [fileGlobs]', 'a comma-separated list of file globs to offline (default: \'**/*\')', '**/*')
   .option('--import-scripts <importScripts>', 'a comma-separated list of additional scripts to import into offline-worker.js')
+  .option('--directory-indexes <directoryIndexes>', 'change files to be a directory only')
   .action(function(dir, options) {
     offline({
       rootDir: dir,
       fileGlobs: options.fileGlobs ? options.fileGlobs.split(',') : null,
       importScripts: options.importScripts ? options.importScripts.split(',') : null,
+      directoryIndexes: options.directoryIndexes ? options.directoryIndexes.split(',') : null,
     })
     .catch(function(err) {
       process.stderr.write(chalk.red.bold(err) + '\n');
--- a/package.json
+++ b/package.json
@@ -37,6 +37,7 @@
     "change-case": "^2.3.0",
     "cli": "^0.11.1",
     "commander": "^2.8.1",
+    "escape-string-regexp": "^1.0.3",
     "fs-extra": "^0.26.0",
     "gh-pages": "https://github.com/marco-c/gh-pages/tarball/817ee16441d72a0c1d33ce109cddf9c1963991d8",
     "git-list-remotes": "^1.0.0",
--- a/templates/app/offline-worker.js
+++ b/templates/app/offline-worker.js
@@ -44,7 +44,6 @@
 
     // This is a list of resources that will be cached.
     RESOURCES: [
-      './', // cache always the current root to make the default page available
 <% resources.forEach(function (pathAndHash) {
 %>      '<%- pathAndHash.path %>', // <%- pathAndHash.hash %>
 <% }); %>


lib/offline.js
```diff
@@ -110,6 +111,15 @@ module.exports = function(config) {
       }, {}));
     }
 
+    function getPathName(filepath, rootDir, directoryIndexes) {
+       filepath = filepath.replace(new RegExp('/(' + directoryIndexes.join('|') + ')$'), '/');
```
mykmelez: This regex constructor call should escape the values of _directoryIndexes_, so the regex only matches their literal values. It may not matter for the common case, because _index.html_ isn't likely to match anything else (although the period is a special character, so it _could_ do so); but it seems like a footgun, likely to fail in unexpected ways that are challenging to diagnose.

The https://www.npmjs.com/package/escape-string-regexp module seems like it'll do the job, and its interface is compatible with _Array.map_, so this should be as simple as `directoryIndexes.map(escapeStringRegexp).join('|')` (although I haven't tested that yet!).

Also, since the regex won't change over the lifetime of the call to the _module.exports_ outer function, you can theoretically optimize by creating the regex once and then reusing it for each call to _getPathName_. Although I'm not sure how much difference that makes in practice, since I think JS engines know how to optimize it away these days.

Either way, you don't need to pass _directoryIndexes_ into _getPathName_, since it has access to that variable via the outer function, whose scope it closes over.

mykmelez:
Because this rewrites at build time (i.e. in the code to generate the service worker), it has a disadvantage over [sw-precache's directoryIndex option](https://github.com/GoogleChrome/sw-precache#directoryindex-string), which handles this issue at runtime (i.e. in the service worker code): it requires a redirect from `/index.html` to `/` in order to ensure that the user will always load the cached version of the file (since `/index.html` is no longer cached).

Without the redirect, the service worker will serve the cached file for `/`, but it won't serve it for `/index.html`; so if the user loads `/index.html`, then they'll always get an uncached version of the file.

Hmm, and thinking a bit more… it might also fail even if there _is_ a redirect, when the device is offline, because the service worker needs to load `/index.html` from the network in order to get redirected to `/`, and that network request will fail without a connection.

So it seems like it'd be better to do this at runtime rather than at build time, regardless of whether we serve `/index.html` in response to `/` (like sw-precache does) or vice-versa (like this branch does).

As for which one to cache, I don't have a strong preference, although it's a bit easier to cache files than directories, given the way _fileGlobs_ works. And there's some precedent for serving `/index.html` in place of `/`, since that's the way web servers work. So it might be more intuitive to folks.

jonathanKingston:
Completely agree with all the comments you gave, I rushed this through to fix the issue with our site but certainly working through this to be better is worthwhile.

The redirect is a edge case that doesn't necessarily ever get called by the service worker so I dunno perhaps giving an option to cache both? But then our site has lots of other redirects that are never cached either, perhaps settings for this might be worth having too?

mykmelez:
> The redirect is a edge case that doesn't necessarily ever get called by the service worker so I dunno perhaps giving an option to cache both?

Indeed, presumably links and references to an app will point to one or the other, probably the `/` variant. Rather than cache both, however, I'd probably cache one and have the service worker (offline-worker.js) map the other one to it, to reduce the cache footprint (so we don't cache the same content twice). This requires us to write the mapping into the service worker when we generate that file, which should be straightforward (the worker is generated via the template in templates/app/offline-worker.js).

> But then our site has lots of other redirects that are never cached either, perhaps settings for this might be worth having too?

Hmm, yes, that does seem useful, if those redirects are likely to be hit. I'd do this as a separate feature, since it's a more general one, and won't be relevant in as many cases.

jonathanKingston:
So I pushed the regex fixes as mentioned however it still caches the directory path rather than the file.
I guess it depends on if you see the cache (Either seem suitable):
- matching the file system on the server
  - Pushing index.html into the cache with rules to serve / with index.html
- matching the messages sent over the wire
  - Pushing / into the cache

The issue I see with the file system mapping is that it makes `oghliner.get` or `oghliner.cachedResources` really quite complex:
- to clone a fake request to match the directory rule
- check the cache for multiple files before looking up using `fetch` as a fallback
- push a file into the cache multiple times

The expansions I wanted to add for the matching the messages over the wire are:
- adding cache pushes to `oghliner.get` to push less important files after the service worker has been activated
- adding simpler ways to limit the globbed files (our site has lots of IE hacks which all get cached causing the service worker to block 2G mobile connection on secondary page loads)

Sorry for pestering, I'm interested in contributing a fair few changes so interested more of the thought process of the project etc.
Thanks for taking time out of your trip in Orlando to respond.

mykmelez:
> So I pushed the regex fixes as mentioned however it still caches the directory path rather than the file.
> I guess it depends on if you see the cache (Either seem suitable):
> 
> ```
> matching the file system on the server
>     Pushing index.html into the cache with rules to serve / with index.html
> matching the messages sent over the wire
>     Pushing / into the cache
> ```

Indeed, either of these are reasonable. I've been thinking of it as matching the filesystem, primarily because we specify the list via file globs. But offline-worker.js undermines that conceptualization when it unconditionally caches `./` in order to "make the default page available."

And ultimately offline-worker.js has no access to the filesystem from which it was served. It only has access to the paths being requested. So it would make sense for its list of resources to be identified by requested path.

> The issue I see with the file system mapping is that it makes oghliner.get or oghliner.cachedResources really quite complex:
> 
> ```
> to clone a fake request to match the directory rule
> check the cache for multiple files before looking up using fetch as a fallback
> push a file into the cache multiple times
> ```

Indeed, it does complicate the worker, although the alternative complicates the code that generates the worker.

> The expansions I wanted to add for the matching the messages over the wire are:
> 
> ```
> adding cache pushes to oghliner.get to push less important files after the service worker has been activated
> adding simpler ways to limit the globbed files (our site has lots of IE hacks which all get cached causing the service worker to block 2G mobile connection on secondary page loads)
> ```

Right, these are valuable use cases to handle. (Related to the first one, there's also the "cache on demand" use case, for files that should be cached only if requested. That might also be another way to handle the IE hacks issue, if you specify the IE files as "cache on demand".)

> Sorry for pestering, I'm interested in contributing a fair few changes so interested more of the thought process of the project etc.

No apology needed! To the contrary, this conversation has revealed aspects of the design that weren't thought through well enough, so it's very useful!

> Thanks for taking time out of your trip in Orlando to respond.

I'm back now, so I should be a bit more responsive! :smile:

README.md
```diff
@@ -190,6 +190,7 @@ The *offline* command takes the following options:
 
 - `--file-globs glob,…` or `fileGlobs: ['glob', …]` - a comma-separated list of globs identifying the files to offline (default: `**/*`). The globs are matched inside *rootDir*.
 - `--import-scripts script,…` or `importScripts: ['script', …]` - a comma-separated list of additional scripts to evaluate in the service worker (no default value). This is useful, for example, when you want to use the [Push API](https://developer.mozilla.org/en-US/docs/Web/API/Push_API).
+- `--directory-indexes index.html` or `directoryIndexes: ['index.html', …]` - a comma-separated list of file names that will be replaced to be served as a directory. This fixes links to `'path/name/'` that don't have the file: `'path/name/index.html'`
```
mykmelez: Nit: since the command-line option takes multiple arguments, the --directory-indexes argument should end with `,…` to indicate that, i.e.:

```
--directory-indexes index.html,…
```

mykmelez:
> So I pushed the regex fixes as mentioned

Hmm, I don't see any additional commits on this branch, and the regex looks the same as it did before.

test/testOffline.js
```diff
@@ -257,6 +257,31 @@ describe('Offline', function() {
...
+    fs.writeFileSync(path.join(subDir, 'test_file_3.js'), 'test_file_3');
+
+    process.chdir(rootDir);
+
+    return offline({
+      rootDir: dir,
+      directoryIndexes: ['test_file_1.js']
+    }).then(function() {
+      var content = fs.readFileSync(path.join(dir, 'offline-worker.js'), 'utf8');
+      assert.notEqual(content.indexOf('/'), -1, 'Must have a directory in the output');
```
mykmelez: This won't actually ensure that there's a directory in the output, since offline-worker.js is full of forward slashes anyway. Looking for `./` would be more reliable.

But that reminds me that offline-worker.js already caches `./` unconditionally. With _directoryIndexes_ set to `['index.html']` and a top-level index.html file, that creates a duplicate entry for `./` in the worker's _RESOURCES_ array.

It seems like we should drop that unconditional entry and instead set the default value of _directoryIndexes_ to `['index.html']`, which seems by far the most likely value (given that it's the default _DirectoryIndex_ for the servers that host GitHub Pages.)

mykmelez:
Aside: I wonder if it would make sense to note files that match a _directoryIndexes_ item, so it's more obvious that those will be served in response to a request for their directory. Currently the output looks like this:

```
> oghliner offline --directory-indexes index.html dist/
Offlining dist/ to dist/offline-worker.js…
…
✓ Caching dist/index.html (3.04 kB)
```

Whereas something like this would be informative and might reduce some confusion:

```
> oghliner offline --directory-indexes index.html dist/
Offlining dist/ to dist/offline-worker.js…
…
✓ Caching dist/index.html at path ./ (3.04 kB)
```

mykmelez:
One last note: it'd still be useful to figure out how to ensure that `./index.html` also gets cached (or gets retrieved from the cache entry for `./`.)

jonathanKingston:
> One last note: it'd still be useful to figure out how to ensure that ./index.html also gets cached (or gets retrieved from the cache entry for ./.)

So I made it push that path into the array of paths for good measure and filter out any duplicates.

> Aside: I wonder if it would make sense to note files that match a directoryIndexes item

Added :+1: 

> Hmm, I don't see any additional commits on this branch, and the regex looks the same as it did before.

Yeah my mistake sorry.

> set the default value of directoryIndexes to ['index.html']

Done :+1: 

> Right, these are valuable use cases to handle. (Related to the first one, there's also the "cache on demand" use case, for files that should be cached only if requested. That might also be another way to handle the IE hacks issue, if you specify the IE files as "cache on demand".)

We actually whitelist file paths to be used in a cache on demand strategy, post install pushes would also be useful.
For IE hacks I have actually stripped the html files to not contain them before running oghliner and forced oghliner to look for `file.sw.html` (See here: https://www.cyber-ami.com/offline-worker.js). This might be something that could be added to oghliner as a compile option potentially however I suspect it could be a complete edge case nobody would use.

Thanks once again!

lib/offline.js
```diff
@@ -119,19 +131,23 @@ module.exports = function(config) {
         var data = fs.readFileSync(filepath);
         totalGzipSize += gzipSize.sync(data);
         var hash = getHash(data);
-        process.stdout.write(chalk.green.bold('✓ ') + 'Caching ' + filepath + ' (' + prettyBytes(size) + ')\n');
+        var path = getPathName(filepath, rootDir);
+
+        process.stdout.write(chalk.green.bold('✓ ') + 'Caching ' + filepath + ' at path ' + path + ' (' + prettyBytes(size) + ')\n');
```
mykmelez: I would only add a reference to the path for files that match a _directoryIndexes_ value (and whose paths are thus being modified); otherwise the reference doesn't stand out, and the output is cluttered by largely-redundant references to paths that resemble their files, f.e.:

```
> oghliner offline --directory-indexes index.html dist/
Offlining dist/ to dist/offline-worker.js…
✓ Caching dist/CNAME at path ./CNAME (13 B)
✓ Caching dist/css/index.css at path ./css/index.css (378 B)
✓ Caching dist/favicon/android-chrome-144x144.png at path ./favicon/android-chrome-144x144.png (2.65 kB)
…
✓ Caching dist/index.css at path ./index.css (223 B)
✓ Caching dist/index.html at path ./ (3.04 kB)
✓ Caching dist/index.js at path ./index.js (4.29 kB)
```

test/testOffline.js
mykmelez: I would still change this line to look for the index of the text `'./'` (including the single quotes) to make it a more reliable assertion about the presence of a reference to the root directory.

jonathanKingston:
@mykmelez adding in the better test made me realise it wasn't working correctly thanks! Changes are pushed for your last comments now.

lib/offline.js
```diff
@@ -119,19 +131,30 @@ module.exports = function(config) {
...
 
-      return filesAndHashes;
+      filesAndHashes.unshift({
+        path: './'
+      }); // cache always the current root to make the default page available
+
+      // Remove duplicates from file list
+      return filesAndHashes.filter(function(fileName, index) {
+        return filesAndHashes.indexOf(fileName) === index;
+      });
```
mykmelez, line 157: This doesn't seem to be working, as I still see two occurrences of `'./'` in the list of _RESOURCES_ when I offline a project using this branch:

>    RESOURCES: [
>      './', // 
>      …
>      './', // 238fefe39dae24af2aa4b62f61f3745953b4858c

It also doesn't seem like the right one would win in this case, since the one without the hash appears first in the list and would presumably be selected, which drops information.

mykmelez, line 157: Given that _directoryIndexes_ now has a default value, we shouldn't need to "cache always the current root," as that value should ensure that it's cached. So I would simply remove that code, which also means that we don't need to do duplicate detection.

mykmelez:
> One last note: it'd still be useful to figure out how to ensure that ./index.html also gets cached (or gets retrieved from the cache entry for ./.)
> 
> So I made it push that path into the array of paths for good measure and filter out any duplicates.

Hmm, I might be missing something, but it looks like ./index.html still doesn't get cached if _directoryIndexes_ includes `index.html`, nor does a request for that file cause the service worker to retrieve the cached entry for `./`. It should do one of those two things, so users who access a page via a URL that includes the name of the file (i.e. https://example.com/index.html) can re-access it after going offline.

jonathanKingston:
> Hmm, I might be missing something, but it looks like ./index.html still doesn't get cached if directoryIndexes includes `index.html`

Perhaps this is a me misunderstanding however I had assumed we had agreed:
- Oghliner was going to match server paths
  - Specifying a `directoryIndex` would tell the browser to cache the directory paths for the matching file system. Files at index.html would likely be a redirect to the directory on live and could safely be ignored as shouldn't be linked to.
- A second piece of work not in this commit was going to include cache on demand
  - This would mop up any issues with links not linking to the directory path

The failed fetch to the `index.html` path could also check the directory path in the cache?

lib/offline.js
jonathanKingston, line 157: This certainly makes sense in a setup where you can specify to prefer the `index.html` over the `/` directory path. However I will await your response on my latest comment which should hopefully clear up the confusion.

mykmelez:
> Perhaps this is a me misunderstanding however I had assumed we had agreed:

Sorry, I probably didn't communicate well myself!

> Oghliner was going to match server paths

Yes, that's the basis for the current approach: the list of _RESOURCES_ in the service worker identifies the set of server (URL) paths (relative to the service worker path) to resources that the web app needs in order to load and run offline, and Oghliner maps files matching _directoryIndexes_ to their server paths at build time.

> Specifying a directoryIndex would tell the browser to cache the directory paths for the matching file system.

Yes, but I've taken this to mean that it'll cache those directory paths _in addition to_ the paths to the files. So the default value of _directoryIndexes_, `['index.html']`, will result in a top-level index.html file being cached at both `./` and `./index.html` (either by caching it twice or, ideally, by mapping from one to the other in the service worker).

> Files at index.html would likely be a redirect to the directory on live and could safely be ignored as shouldn't be linked to.

I don't think we can assume this. For GitHub Pages, which is the deployment target that Oghliner was originally developed to support (although it isn't required to use the offlining features of Oghliner), it definitely isn't the case, as GitHub Pages doesn't do this redirection, nor is it possible to configure it to do so.

jonathanKingston:
> either by caching it twice or, ideally, by mapping from one to the other in the service worker

Yeah I don't want to wait around for another cache certainly - I was trying to avoid that. I will get the service worker rerouteing those I think.
I'll get that added as part of this work thank you.

> I don't think we can assume this. For GitHub Pages, which is the deployment target that Oghliner was originally developed to support

Certainly I want to keep this the primary use-case as even our use case is the same using a CDN. I also don't want to assume the redirection either so I guess that could be something that is configured later.

It does raise an interesting point however, in that perhaps the core of oghliner could be separated out for more complex use-cases leaving oghliner a github setup and configuration of the separate core lib. (This would be long term goal however I am interested in different caching strategies and also interrogating the cache for things like 'suggested products' sections which would load items that are within the cache). I'm not sure if this is the most ideal Mozilla extension point for this kind of work however it would be something I am interested in working with you on.

mykmelez:
> It does raise an interesting point however, in that perhaps the core of oghliner could be separated out for more complex use-cases leaving oghliner a github setup and configuration of the separate core lib. (This would be long term goal however I am interested in different caching strategies and also interrogating the cache for things like 'suggested products' sections which would load items that are within the cache). I'm not sure if this is the most ideal Mozilla extension point for this kind of work however it would be something I am interested in working with you on.

We've discussed doing something like this, and @delapuente did some work to integrate https://github.com/delapuente/offliner, which provides an application lifecycle that is separate from the service worker lifecycle, into Oghliner over in #96, although we ultimately decided not to take that integration for Oghliner 1.0.

https://github.com/delapuente/offliner is designed to work with https://github.com/fxos-components/serviceworkerware, which provides a generic framework for writing components that run in a service worker. Thus it's more akin to Google's https://github.com/GoogleChrome/sw-toolbox (except that ServiceWorkerWare is more modular), while Oghliner is more similar to Google's https://github.com/GoogleChrome/sw-precache (plus support for deployment to GitHub Pages).

My team at Mozilla is considering a project to continue development of ServiceWorkerWare in some way over in https://github.com/mozilla/wadi/issues/33, and that issue provides a useful comparison of the similarities and differences between ServiceWorkerWare and sw-toolbox. I recommend taking a look at ServiceWorkerWare, as it might be the best Mozilla tool for the work you're looking to do with service workers.

Having said that, however, I'm also open to integrating more advanced functionality into Oghliner. It already supports _importScripts_ to import arbitrary scripts into the service worker, and I can imagine a v2 that includes support for other ways to integrate more complex functionality, including ServiceWorkerWare components like offliner.

As long as we keep the experience simple for the basic use case that Oghliner was originally developed to support, deploying offline web apps to GitHub Pages, I'm happy to add enhancements for more sophisticated/complex use cases.

delapuente:
@jonathanKingston correct me if I'm wrong. You are trying to add a feature that takes a list of files: `['/path/to/file.html','/path/to/another/different-file.html']` so if they are found in the list of files to be cached, they will be replaced by `/path/to/` and `/path/to/another/` in the resource list so the SW won't add those files but the directories. Isn't it?

Imagine. With this folder structure:

```
dist
├── index.html
└── models
    └── all.html
```

Now if I specify:

``` js
fileGlobs: [
    '**/*.html'
],
```

It becomes this resource list:

``` js
RESOURCES: [
  './',
  './index.html',
  './models/all.html'
];
```

But with your patch and specifying:

``` js
fileIndexes: [
  'index.html',
  'all.html'
]
```

it would become:

``` js
RESOURCES: [
  './',
  './models/'
];
```

Is that correct?

jonathanKingston:
@mykmelez thanks for pointing me to https://github.com/mozilla/wadi so much to learn!
Certainly seems to make sense to align the internals of oghliner with the other projects mentioned as you say.

@delapuente that's exactly what I am doing. I'm working on making the service worker serve both directory and matching files based upon @mykmelez's feedback.

delapuente:
In issue #253 we have an specific use case asking for a feature like that supports the main goal of providing integration with gh-pages. I'm going to provide a specific solution for it but I would like to progress with this PR and offer a flexible enough solution to solve #253 and @jonathanKingston use case.

From reviewing the discussion, here is the plan I came out with:
1. Remove the forced `./` entry
2. Provide @jonathanKingston proposed `--directory-indices` option as a set of globs. It defaults into `{ "./**/index.html": true }`
3. In the generation step, extend the globs:
   1. Ensure these resources will be added to `RESOURCES` list.
   2. Normalize the set into a map of _cache lookup redirections_ (i.e. `{ "./": "./index.html", "./features/network-indepndent/": "./features/network-independent/index.html" }` )
4. In order to reduce cache fingerprint, treat this feature dynamically in the SW: if the request URL matches one of the redirections, use the value as cache lookup and serve the response.

In sight problems: it is possible the SW should be aware of the site's base URL to properly handle the redirections. We could provide an option `--baseurl` but, at this moment, I think it suffices to assume the SW will be always at the root of the site and take the script's folder URL as the base URL for the entire site.

In the future, if needed, we could turn the _cache lookup redirection_ map into an option.
@jonathanKingston do you think this solution still solves your use case?
@mykmelez can you provide some feedback for the plan?

jonathanKingston:
@delapuente sorry I have not updated the code here (still away from a decent dev setup right now).
Your proposed solution looks like what I was going to end up with besides some semantic detail etc but yes it seems like it will solve my use-case and be extensible enough to add in my own redirects later etc.

Happy to work on this if you are very busy etc?

delapuente:
Let's work in parallel. I'm going to provide the specific support for gh-pages while you can provide this more general option. But before, let's wait for @mykmelez feedback. I don't want to be the killjoy but maybe the feature is beyond oghliner purposes.

jonathanKingston:
@delapuente perfectly happy to believe oghliner is better focussed on simpler use cases. Hence I'm not really pushing hard for either:
- different content for offline vs normal
- redirections

However when it comes to fxos apps which I will be working on a whole lot more soon I would like a lot more tooling like this.

marco-c:
I think Oghliner should be focused on simple use cases, but be configurable to also support more complex ones (although not as complex as writing a service worker yourself).

mykmelez:
> @mykmelez can you provide some feedback for the plan?

That sounds like a great plan!

> I think Oghliner should be focused on simple use cases, but be configurable to also support more complex ones (although not as complex as writing a service worker yourself).

Agreed!

I do think there's room for something in the middle, between Oghliner and a completely hand-written service worker, to best serve more complex needs. Maybe that's https://github.com/GoogleChrome/sw-toolbox (it doesn't have to be a Mozilla project!), but maybe it's https://github.com/fxos-components/serviceworkerware plus https://github.com/delapuente/offliner.

marco-c:
#254 landed, so this PR should now be only about making the stuff from #254 configurable.

mykmelez:
This appears to have stalled, so I'll close it. But please reopen it if you continue working on it!

PR #236, opened 2015-12-19
Adding in ability to specify offline directory.

jonathanKingston:
This needs further review (it's late here and I wanted to push so I don't forget it) however I thought @mykmelez you might be interested in looking at it.

The idea is to add support for the service worker to cache all the `RESOURCES` files with a prefix. This is useful to filter out all the code that browsers that don't support service worker need/have.

This is a similar concept we are using on our site but will also allow for neater stylesheets also - the idea was to reduce that initial resources load down as much as possible to get the finite functional set to read the site. This resource files actually cause network thrashing on mobile devices which will mostly not be an issue with http/2 however changing content to be lighter should always benefit the user.

### DIFF
--- a/README.md
+++ b/README.md
@@ -190,6 +190,7 @@ The *offline* command takes the following options:
 
 - `--file-globs glob,…` or `fileGlobs: ['glob', …]` - a comma-separated list of globs identifying the files to offline (default: `**/*`). The globs are matched inside *rootDir*.
 - `--import-scripts script,…` or `importScripts: ['script', …]` - a comma-separated list of additional scripts to evaluate in the service worker (no default value). This is useful, for example, when you want to use the [Push API](https://developer.mozilla.org/en-US/docs/Web/API/Push_API).
+- `--prefixResourceDirectory` or `prefixResourceDirectory: '/sw'` - a path prefix which is used by the service woker when it fetches the resources on the client. (default: `''`) This is useful to provide different content for offline users.
 
 ## Deploy
 
--- a/lib/offline.js
+++ b/lib/offline.js
@@ -51,6 +51,7 @@ module.exports = function(config) {
 
     var fileGlobs = config.fileGlobs || ['**/*'];
     var importScripts = config.importScripts || [];
+    var prefixResourceDirectory = config.prefixResourceDirectory || '';
 
     // Remove the existing service worker, if any, so the worker doesn't include
     // itself in the list of files to cache.
@@ -86,6 +87,7 @@ module.exports = function(config) {
         cacheId: cacheId,
         cacheVersion: getHashOfHashes(pluckHashes(filesAndHashes)),
         resources: filesAndHashes,
+        prefixResourceDirectory: prefixResourceDirectory,
         importScripts: importScripts,
       };
 
--- a/templates/app/offline-worker.js
+++ b/templates/app/offline-worker.js
@@ -50,6 +50,9 @@
 <% }); %>
     ],
 
+    // This is where the service worker will call for resources, allowing for a smaller service worker version of the same file.
+    prefixResourceDirectory: '<%= prefixResourceDirectory %>',
+
     // Adds the resources to the cache controlled by this worker.
     cacheResources: function () {
       var now = Date.now();
@@ -61,6 +64,9 @@
           var url = new URL(resource, baseUrl);
           var bustParameter = (url.search ? '&' : '') + '__bust=' + now;
           var bustedUrl = new URL(url.toString());
+          if (this.prefixResourceDirectory) {
+            bustedUrl.pathname = this.prefixResourceDirectory + bustedUrl.pathname;
+          }
           bustedUrl.search += bustParameter;
 
           // But cache the response for the original request
--- a/test/testOffline.js
+++ b/test/testOffline.js
@@ -53,6 +53,18 @@ describe('Offline', function() {
 
     return offline({
       rootDir: dir,
+      prefixResourceDirectory: '/sw'
+    }).then(function() {
+      var content = fs.readFileSync(path.join(dir, 'offline-worker.js'), 'utf8');
+      assert.notEqual(content.indexOf('\'/sw\','), -1);
+    });
+  });
+
+  it('should create offline-worker.js in the destination directory', function() {
+    var dir = temp.mkdirSync('oghliner');
+
+    return offline({
+      rootDir: dir,
     }).then(function() {
       assert.doesNotThrow(fs.accessSync.bind(fs, path.join(dir, 'offline-worker.js')));
     });


mykmelez:
I'd be interested in seeing an example of this! It would help me understand the use case better. Is the code for your site available?

jonathanKingston:
So the code for my current work is available here: https://www.cyber-ami.com/offline-worker.js (I'm not sure on the rights to reuse for this) however this is replacing `index.html` files with `index.sw.html` such that non service worker browser gets: https://www.cyber-ami.com/index.html and service worker loads: https://www.cyber-ami.com/index.sw.html

The principle is the same in my solution proposed here however using a directory should be easier to manage.

mykmelez:
Aha, thanks! That raises another question!

Is there something specific about the way that the service worker loads files that causes the network thrashing? I would have thought that you'd want to serve the "lighter" version of the site for all mobile devices, regardless of whether or not they support service workers, since any issue with network thrashing would happen regardless of whether the site was being loaded by a service worker or by a mobile browser that doesn't support them. Is it because the service worker retrieves all the files used by the site, which is a superset of the files that are used by the first page of the site?

jonathanKingston:
> Is it because the service worker retrieves all the files used by the site, which is a superset of the files that are used by the first page of the site?

Initially this was the case however I managed to hack ogliner output so that it only contains the files we care about and lazily push files in afterwards.

For example on a slow 2G connection oghliner starts installing the service worker and creates 10 open connections to the site which blocks the current interaction of the site as I as the user get locked out from using any more connections. Our site on a 2G connection was taking 5-10 seconds to download which lead to 10 seconds of broken interface to the user.

Reducing the number of files helped but also we get other advantages from loading custom HTML for service worker like for example:
- cut out critical CSS as that has loaded
- I preload the fonts and don't swap them out with JavaScript on DOM load (This causes a notable flicker on non service worker clients that I wanted to prevent) so the HTML has an extra class in to use our custom font rather than depend on JS to add that in.
- Shims are all removed as they are not needed or flex hacks using Modernizr

> I would have thought that you'd want to serve the "lighter" version of the site for all mobile devices

Ideally yes we would, however we also want to support older browsers and I don't really want to do anything more on the server for checking for browser strings to serve the correct HTML.

mykmelez:
> Ideally yes we would, however we also want to support older browsers and I don't really want to do anything more on the server for checking for browser strings to serve the correct HTML.

Ah, that makes sense. You're using "supports service workers" as a simple way to distinguish between "modern browsers," which have both service workers and a variety of other modern features, and "other browsers," some of which may have other modern features, but which you lump together with older browsers for simplicity.

So you generate two sets of HTML files, one for modern browsers and one for other browsers. Both sets of files reference the same URLs, so that it's easier to reuse code between the two versions of the site. And the service worker dynamically rewrites the URLs that it loads to point to the service worker-specific files.

There's one thing that confuses me about using this technique to reduce network thrashing. The first time the site loads, it will load the non-service-worker versions of the files (for those that are loaded during page load, anyway), since the service worker won't have been installed/activated yet.

When the service worker gets installed and starts caching resources, if it requests a different set of URLs than the ones that the page has already loaded, then I would expect that to increase network usage, since the service worker will cause all of those resources to be retrieved anew. Whereas if it was requesting the same set of URLs as the ones that the page has already loaded, then presumably it would retrieve some of those from the browser's memory/disk caches.

Hmm, except I just remembered that the service worker adds a parameter to each URL to bypass/bust the browser caches, so that it always retrieves the resources anew. I wonder if we should stop doing that (at least the first time a site is loaded, i.e. on the initial install of a service worker), given that it can effectively double network usage. That might address the network thrashing issue.

Regardless, the other use case is to serve a different version of the site to modern browsers. For that use case, I have two questions:
1. Is it ok that modern web browsers will load the non-service worker version of the site the first time the site is loaded?
   Presumably you'd want them to get the same version on first-load that they get on subsequent loads, if there's meaningful difference in the functionality of the two versions. I'm not sure how to accomplish that with dynamic URL rewriting in the service worker, however, since at least some of the site (if not most/all of it) will load before the service worker gets activated.
   I suppose index.html could be a "shell" that first installs the service worker and then loads the site once it knows that the service worker is activated and will be intercepting requests. But that would probably slow down first load too much.
2. Is a single option that modifies all URLs fine-grained enough for this use case?
   Presumably many resources are still the same between the two versions of the site (f.e. images), even if the HTML files and some other resources are different. If the service worker rewrites the URLs for all resources, then there needs to be two copies of them on the server, even for the ones that don't differ.
   It looks like the service worker on your site currently only rewrites the URLs for HTML files, whereas _prefixResourceDirectory_ would rewrite the URLs for all resources. Which seems like it'd make you copy a bunch of files that don't actually differ between the two versions.

jonathanKingston:
> There's one thing that confuses me about using this technique to reduce network thrashing.

The size of the files are much smaller after removing the fat for much older browsers we can actually cut even more also.
So we end up cutting the number of packets needed to get all the HTML causing the network blocking but also the number of secondary requests.

> Hmm, except I just remembered that the service worker adds a parameter to each URL to bypass/bust the browser caches

Yup I couldn't find a sensible way to turn that off. I'm also not that familiar with when a browser decides to bust it's own cache. Not loading resources for IE CSS fixes seemed sensible on second load even if just from a browser memory perspective.

> Is it ok that modern web browsers will load the non-service worker version of the site the first time the site is loaded? 

No, the intention is just to cut out cruft not add any new features. The functionality should be exactly the same in offline mode.
The shell behaviour you describe is something I have been experimenting with though to give SPA feel and reduce even further the HTML downloaded. Even this behaviour is built in a way that just loads after the page renders for non sw browsers.

> Is a single option that modifies all URLs fine-grained enough for this use case?

It would be for us as the duplication of storing a few mb of files is not an issue. However I can see how that would be an issue for others (This is why I wanted the discussion regarding this sample code - thanks!).
We would however benefit from dialling down our cssnext settings to only be browsers that support service worker which would get us further reduced CSS size.
Perhaps a glob based file check to flip between using `prefixResourceDirectory` and not would be much more flexible?

mykmelez:
> No, the intention is just to cut out cruft not add any new features. The functionality should be exactly the same in offline mode.

Hmm, I'm not sure how to guarantee that first load is the same as second load for a browser that supports service workers. At the very least, the initial HTML page will be the non-service worker version (i.e. /index.html).

Perhaps that page could check if the browser supports service workers and, if so, immediately redirect to the service worker version of the page (i.e. /sw/index.html or index.sw.html), before the service worker is even registered, possibly even before the page loads, f.e. in a script within the <head> section of the page.

> Perhaps a glob based file check to flip between using prefixResourceDirectory and not would be much more flexible?

Hmm, yes, I think that would be more flexible, although then I think you'd need two options: a glob/globset to specify the set of paths to match, and the prefix the prepend to paths that match.

I wonder how much trouble it would be to rewrite the paths in the service worker version of the page to include the sw/ subdirectory. Then the service worker wouldn't have to transform any paths at all.

Alternately, if the paths are all relative, and the user gets directed to the service worker version of the page on first load, then perhaps you could avoid rewriting paths entirely. For example, if /sw/index.html contains a reference to style/index.css, and the service worker at /sw/offline-worker.js fetches that relative URL, then it should fetch the file at /sw/style/index.css, without having to explicitly rewrite the URL.

jonathanKingston:
> Hmm, I'm not sure how to guarantee that first load is the same as second load for a browser that supports service workers. At the very least, the initial HTML page will be the non-service worker version (i.e. /index.html).

This is down to the implementer to get right they should only transform code for performance.

> Perhaps that page could check if the browser supports service workers and, if so, immediately redirect to the service worker version of the page

I'm using it as a cut the mustard test but also 'has cache loaded' improvements. So redirecting those users would have a negative impact as they would load a page without CSS in the `<head>` at all including critical CSS.

I'm thinking renaming extensions based upon a glob is the least confusing when we are starting to talk about limiting the extensions due to the relative path rewriting. Perhaps worth experimenting with both.

I'm going to be offline for about a week but will pick up the examples when I get back.

mykmelez:
> Hmm, I'm not sure how to guarantee that first load is the same as second load for a browser that supports service workers.

The Stack Overflow thread [Use Service Worker to intercept requests on the first load of my site?](http://stackoverflow.com/questions/34681583/use-service-worker-to-intercept-requests-on-the-first-load-of-my-site) suggests an approach.

jonathanKingston:
@mykmelez I would have to have a play about with using that solution. When I used it I was getting tonnes of broken cache requests. If it would work that way it would solve a lot of the issues we had regarding loading the content twice and the cache busting issues etc.

I'm back from holiday now so I can give this a whole heap more focus :D.

mykmelez:
This appears to have stalled, so I'll close it. But please reopen it if you continue working on it!

PR #237, opened 2015-12-21
Warn (instead of failing) if a file in importScripts doesn't exist or is a directory

marco-c:


lib/offline.js
```diff
@@ -136,7 +136,14 @@ module.exports = function(config) {
 
     function checkImports(imports, rootDir) {
       imports.forEach(function (filepath) {
-        assertExists(path.join(rootDir, filepath));
+        try {
+          assertExists(path.join(rootDir, filepath));
+        } catch (ex) {
+          process.stdout.write(chalk.yellow.bold(
+            '⚠ ' + filepath + ' doesn\'t exist or is a directory, but you\'ve added it to importScripts.\n' +
+            '  Are you sure you want to import it?\n'
```
mykmelez: This asks a question but doesn't prompt the user to answer it. I wouldn't actually prompt here, since the user may have a good reason for doing this (f.e. they intend to create the file but haven't done so yet). And it's a warning, not an error (thus doesn't need to be corrected). But I'd either direct the user to address the problem:

> ⚠ file/path doesn't exist or is a directory, but you've added it to importScripts.
>  To fix this problem, ensure file/path is a file or remove it from importScripts.

Or simply note the problem without any direction:

> ⚠ file/path doesn't exist or is a directory, but you've added it to importScripts.

lib/offline.js
```diff
@@ -149,12 +156,10 @@ module.exports = function(config) {
       try {
         stat = fs.statSync(filepath);
       } catch (ex) {
-        process.stdout.write(chalk.red.bold(filepath + ' doesn\'t exist.'));
         throw ex;
       }
```
mykmelez: Without the _process.stdout.write_ call, there's no reason to catch the exception that _fs.statSync_ throws if the path doesn't exist, so this should simply call fs.statSync and let the exception propagate automatically.

But once you do that, there's little value to factoring out the two checks into _assertExists_ at all, since it only has one caller and is only four lines long:

```
function assertExists() {
  var stat = fs.statSync(filepath);
  if (!stat.isFile()) {
    throw new Error(filepath + ' is not a file.');
  }
}
```

So I would unrefactor it and move its logic into _checkImports_. I would then also continue to distinguish between the two possible error states instead of lumping them together into `doesn't exist or is a directory`, since we know which one is the case, and a more-specific error message will be less confusing.

lib/offline.js
```diff
@@ -135,8 +135,23 @@ module.exports = function(config) {
...
+            warning = 'is a directory';
+          }
+        } catch (ex) {
+          warning = 'doesn\'t exist';
+        }
+
+        if (warning) {
+          process.stdout.write(chalk.yellow.bold(
+            '⚠ ' + filepath + ' ' + warning + ', but you\'ve added it to importScripts.\n' +
+            '  To fix this problem, ensure file/path is a file or remove it from importScripts.\n'
```
mykmelez: Erm, when I wrote _file/path_ in my suggested warning messages, I meant for that to be a placeholder for the actual file path!

However, in retrospect, it's probably overkill to include the file path twice in this message. So I would simply change "ensure file/path is a file" to "ensure the path is a file."

marco-c: I don't know how I didn't notice the "file/path" string...

Fixed in e2e024d.

PR #238, opened 2015-12-21
Don't interrupt spinner on errors not handled by handleCredentialErrors

marco-c:


marco-c:
There's a timeout when the tests are run via `npm test` or `gulp test`, which doesn't happen if you directly run the tests using `mocha`.

marco-c:
Fixed in 2a7b736.

mykmelez:
> There's a timeout when the tests are run via npm test or gulp test, which doesn't happen if you directly run the tests using mocha.

Hmm, is this a race condition? We may want 2a7b736 regardless, I'm just trying to understand why this only surfaced now and if that change papers over an underlying issue.

marco-c:
No, it wasn't a race condition, but a regression caused by not interrupting the spinner in `handleCredentialErrors` for errors not related to credentials (that is, a0a53043d2f578645d92fffc003e388ae061ba46).

PR #241, opened 2015-12-21
Show options in the .gh-pages-cache prompt and set Y as the default

### LINKED ISSUE #235: improve .gh-pages-cache confirm prompt
The confirm prompt that asks if you want to add .gh-pages-cache to your .gitignore file doesn't specify a default option. It should specify one, probably "yes".

It's also very long, prone to wrapping in wide terminals, which is fine, except that it's inconsistent with other long strings, which we explicitly wrap. We should wrap this one too.

I might also add some additional explanatory text and make it a warning (complete with an icon), so it becomes something like:

```
⚠ The .gh-pages-cache directory is a temporary repository that I use
  to push changes to your gh-pages branch. It isn't in your .gitignore file,
  which means that Git will report it as an untracked file.

Do you want to add .gh-pages-cache to .gitignore (Y/n)?
```

marco-c:
Fixes #235.

### DIFF
--- a/cli.js
+++ b/cli.js
@@ -74,11 +74,13 @@ program
         var gitignore = fs.readFileSync('.gitignore', 'utf8');
 
         if (gitignore.indexOf('.gh-pages-cache') === -1) {
-          return promptly.confirm(chalk.yellow.bold(
+          process.stdout.write(chalk.yellow.bold(
             '⚠ .gh-pages-cache is a temporary repository that I use to push\n' +
             '  changes to your gh-pages branch. It isn\'t in your .gitignore file,\n' +
-            '  which means that Git will report it as an untracked file.\n\n' +
-            '  Do you want to add .gh-pages-cache to .gitignore (Y/n)?'), { default: true })
+            '  which means that Git will report it as an untracked file.\n\n'
+          ));
+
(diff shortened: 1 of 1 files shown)


cli.js
```diff
@@ -74,7 +74,12 @@ program
         var gitignore = fs.readFileSync('.gitignore', 'utf8');
 
         if (gitignore.indexOf('.gh-pages-cache') === -1) {
-          return promptly.confirm('.gh-pages-cache is a temporary repository that we use to push changes to your gh-pages branch. Do you want to add it to .gitignore?').then(function(answer) {
+          return promptly.confirm(chalk.yellow.bold(
+            '⚠ .gh-pages-cache is a temporary repository that I use to push\n' +
+            '  changes to your gh-pages branch. It isn\'t in your .gitignore file,\n' +
+            '  which means that Git will report it as an untracked file.\n\n' +
+            '  Do you want to add .gh-pages-cache to .gitignore (Y/n)?'), { default: true })
```
mykmelez: Nit: I would align this line with the alert icon, not with the alert text, and make it the default text color rather than the alert color, as I see it as logically separate from the alert itself. That is: it prompts the user to do something about the issue identified by the alert.

marco-c: I tried both and couldn't decide which one was better, I'll follow your advice: 3e97a2f24d1ed8638e919243e192cd6968f41c76

PR #244, opened 2015-12-23
Don't reinstall oghliner to have its development version but directly install it during bootstrap

marco-c:
Looks like newer versions of `npm` don't install the `gulp` binary in `node_modules/.bin/gulp` if `gulp` is a sub-dependency (but in `node_modules/subdependency/.bin/gulp`).


marco-c:
Actually, it was installed by `bootstrap` but somehow the newer `npm` version was removing it when reinstalling `oghliner`.

PR #254, opened 2016-01-07
Make requests to directories become requests to index.html

### LINKED ISSUE #253: For each `index.html`, cache its parent directory too
As the common case in the Internet, gh-pages serves `index.html` when the URL points to a directory. This is used by some tools such as Jekyll. In order to fully support this feature, each time an `index.html` file is found, it's parent directory should be cached as well.

delapuente:
Fix #253

### DIFF
--- a/templates/app/offline-worker.js
+++ b/templates/app/offline-worker.js
@@ -109,11 +109,13 @@
 
     // Make requests to directories become requests to index.html
     extendToIndex: function (request) {
-      var url = request.url;
-      if (url[url.length - 1] === '/') {
-        url += 'index.html';
+      var url = new URL(request.url, self.location);
+      var path = url.pathname;
+      if (!path[path.length - 1] === '/') {
+        return request;
       }
-      return new Request(url, request);
+      url.pathname += 'index.html';
+      return new Request(url.toString(), request);
     },
 
     // Prepare the cache for installation, deleting it before if it already exists.
--- a/templates/app/offline-worker.js
+++ b/templates/app/offline-worker.js
@@ -111,7 +111,7 @@
     extendToIndex: function (request) {
       var url = new URL(request.url, self.location);
       var path = url.pathname;
-      if (!path[path.length - 1] === '/') {
+      if (path[path.length - 1] !== '/') {
         return request;
       }
       url.pathname += 'index.html';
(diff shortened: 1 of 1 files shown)


marco-c:
The patch looks good to me, but I'd make this configurable.
Were you planning to make it configurable in a follow-up?

delapuente:
I was not planning to make it configurable as it is not configurable in gh-pages (it is always this way) but we can add a follow up and make it optional.

marco-c:
Sounds good to me.
Let's see what @mykmelez thinks, since he's been reviewing #234.

templates/app/offline-worker.js
```diff
@@ -107,6 +107,15 @@
       });
     },
 
+    // Make requests to directories become requests to index.html
+    extendToIndex: function (request) {
+      var url = request.url;
+      if (url[url.length - 1] === '/') {
+        url += 'index.html';
+      }
+      return new Request(url, request);
```
marco-c: Nit: I guess it doesn't really matter, but we can avoid creating a new Request object when we don't need to rewrite the url.

mykmelez: It might also be worth isolating the pathname within the URL string via something like `new URL(request.url).pathname`, since a URL to a file hosted by GitHub Pages can end in a search string like `?foo=bar` (even if the server doesn't actually process the string) and/or a fragment identifier like `#anchor` (although I'm not sure if that gets preserved by the _Request_ object).

mykmelez:
> Let's see what @mykmelez thinks, since he's been reviewing #234.

I'm ok with this, as it's a useful, GitHub Pages-specific fix, while #234 is a more general solution for folks who reuse Oghliner on other hosts. And it sounds like @delapuente and @jonathanKingston are working together to move #234 forward as well.

templates/app/offline-worker.js
delapuente: Wow. You're totally right here. Nice catch!

delapuente:
@mykmelez @marco-c I added your comments. Do you mind to take a second look?

templates/app/offline-worker.js
```diff
@@ -107,6 +107,17 @@
       });
     },
 
+    // Make requests to directories become requests to index.html
+    extendToIndex: function (request) {
+      var url = new URL(request.url, self.location);
+      var path = url.pathname;
+      if (!path[path.length - 1] === '/') {
```
marco-c: Nit: can you use !== instead?

PR #257, opened 2016-01-11
Adding the infrastructure for enabling browser tests for the service worker.

delapuente:


delapuente:
@marco-c this is some boilerplate needed to start writing the tests. Do you mind to review? Asking @mykmelez for additional input.

I'm not adding these tests to the `test` task as it is not possible to run them without the browsers, so, AFAIK, they can not be executed on Travis but at least, we could test the worker inside the Service Worker environment.

Notice that enabling testing the service worker without browsers is still possible if we _browserify_ the serviceworker and export [`oghliner` API](https://github.com/mozilla/oghliner/blob/master/templates/app/offline-worker.js#L35), mocking everything else.

Actually tests for this interface is exactly what I'm going to do but I prefer to run them inside the real context to be as realistic as possible.

gulpfile.js
```diff
@@ -74,6 +76,23 @@ gulp.task('serve', function () {
   });
 });
 
+gulp.task('sw-test', function () {
```
mykmelez: Nit: I would call this _test-sw_ so the task name starts with a verb.

mykmelez:
> I'm not adding these tests to the test task as it is not possible to run them without the browsers, so, AFAIK, they can not be executed on Travis but at least, we could test the worker inside the Service Worker environment.

It's ok to start with a separate task that we can run locally, but it seems like it should be possible to run these tests on Travis too. For [PluotSorbet](https://github.com/mozilla/pluotsorbet), we use [SlimerJS](http://slimerjs.org/) to run tests in Travis on Firefox/Gecko, and presumably [PhantomJS](http://phantomjs.org/) works similarly for testing on WebKit (although that's not the same thing as testing in Chrome).

Also, Firefox and Chrome themselves are installed in Travis environments and can be run via xvfb. [GUI and Headless Browser Testing](https://docs.travis-ci.com/user/gui-and-headless-browsers/) talks about browser testing on Travis generally, while [How to run JavaScript tests in Chrome on Travis](http://swizec.com/blog/how-to-run-javascript-tests-in-chrome-on-travis/swizec/6647) describes a particular example.

Of course we have to make sure we have a version of each browser that supports service workers. And I'm unfamiliar with Karma, so I don't know how easy it is to hook it up. It's fine to start with tests that just run locally. However, ultimately we should enhance those tests to run in Travis too, and it seems possible to do.

delapuente:
You're completely right. @marco-c was enlightening me about ways of testing browsers on Travis. Waiting for his feedback as well.

package.json
```diff
@@ -33,6 +33,7 @@
     "LICENSE"
   ],
   "dependencies": {
+    "chai": "^3.4.1",
```
marco-c: These dependencies (`chai`, `karma*` and `sinon`) shouldn't be in `dependencies`.

marco-c:
We talked on IRC about SlimerJS and Selenium, but we'll probably be able to directly use Karma on Travis, we just need to download the browsers and maybe create custom karma launchers (they're pretty simple, for example: https://github.com/karma-runner/karma-firefox-launcher/blob/master/index.js).

Here's some documentation: http://karma-runner.github.io/0.8/plus/Travis-CI.html

marco-c:
We'll need to use a custom launcher also because that one doesn't really work on Linux (it always picks up the globally installed Firefox, it doesn't care if it's Nightly, Aurora, etc.).

delapuente:
Don't worry about the launchers, we must override binary paths for the browser via environment variables. It is a known issue I already experimented when working with Arch Linux.

Issue #213, opened 2015-11-19
Test the service worker


delapuente:
We should provide tests for testing the service worker in different platforms.

PR #259, opened 2016-01-12
Testing cacheResources()

delapuente:
@marco-c do you mind to take a look? It's not complete but if you give me feedback now, I can progressively add more tests (a function per commit) and I can fix style errors ASAP.


gulpfile.js
```diff
@@ -94,7 +94,7 @@ gulp.task('test-sw', function () {
 });
 
 gulp.task('pre-test', function () {
-  return gulp.src(['lib/**/*.js'])
+  return gulp.src(['lib/**/*.js,!lib/service-worker/**'])
```
marco-c: This path doesn't exist, does it?

marco-c: Maybe you haven't `git add`ed it?

delapuente: That's not even correct. Let me see what happen there.

delapuente:
@marco-c I finished with the tests. All the API is covered. If you don't mind to take a second look. :)

delapuente:
Fix #213

test/service-worker/testServiceWorker.js
```diff
@@ -0,0 +1,309 @@
...
+  describe('cacheResources()', function () {
+
+    var mockedOkResponse = { ok: true };
+    var mockedFailingResponse = { ok: false };
+    var testTimestamp = 1234567890;
+
+    function allResourcesInCache(resources, cache) {
+      var requests = [];
+      var responses = [];
+      for (var call = 0, entry; entry = cache.put.args[call]; call++) {
```
marco-c: Nit: forEach

test/service-worker/testServiceWorker.js
```diff
@@ -0,0 +1,309 @@
...
+
+        assert(sameRequestURL(resultUrl, originalUrl));
+      });
+    });
+  });
+
+  describe('prepareCache()', function () {
+    function createNonEmptyCache(name) {
+      return self.caches.open(name)
+        .then(function (cache) {
```
marco-c: Nit: for promises, we're using the Q-style in most of the code, e.g.:

``` Javascript
  return self.caches.open(name)
  .then(function(cache) {
    // ...
  });
```

It doesn't really matter though.

marco-c:
Looks good to me.
What do you think about splitting the testServiceWorker.js file in multiple files? (one for each function)

delapuente:
I will fix all the nits and I agree on splitting the test. Maybe not one per method but some of them would be more readable if they were in their own files.

PR #264, opened 2016-01-20
Use arrow functions in the service worker

marco-c:


marco-c:
Ah, tests are failing on Node.js v0.12 because it doesn't support arrow functions.

templates/app/offline-worker.js
```diff
@@ -54,52 +56,44 @@
...
+          }
+          console.error('Error fetching ' + url + ', status was ' + response.status);
+        });
+      })));
     },
 
     // Remove the offline caches not controlled by this worker.
     clearOtherCaches: function () {
-      var deleteIfNotCurrent = function (cacheName) {
+      return self.caches.keys()
```
delapuente, line 84: I would prefer to keep the name as I think it adds readability. We can transform it into an arrow function but keeping the name:

``` js
var deleteIfNotCurrent = cacheName => {
  if (cacheName.indexOf(this.CACHE_PREFIX !== 0 || cacheName === this.CACHE_NAME) {
    return Promise.resolve();
  }
  return self.caches.delete(cacheName);
};

return self.caches.keys()
.then(keys => Promise.all(keys.map(deleteIfNotCurrent)));
```

marco-c, line 84: I originally did this, then made it an anonymous function and added a comment.
I can add it back if you prefer.

test/service-worker/testCacheResources.js
```diff
@@ -29,14 +27,17 @@ describe('cacheResources()', function () {
 
     self.oghliner.RESOURCES = [];
     sinon.stub(Date, 'now').returns(testTimestamp);
-    sinon.stub(self, 'fetch').returns(Promise.resolve(mockedOkResponse));
-    sinon.stub(oghliner, 'prepareCache').returns(Promise.resolve(mockCache));
+    sinon.stub(self, 'fetch', function(request) {
```
delapuente: These changes are not related to this PR. Do you mind to drop them? We can discuss them in another PR.

marco-c: Removed it.

templates/app/offline-worker.js
delapuente, line 84: I missed the comment. Maybe we can write:

``` js
.then(cacheNames => Promise.all(cacheNames.map(cacheName => {
  var isAPreviousCache = cacheName.startsWith(this.CACHE_PREFIX) && cacheName !== this.CACHE_NAME;
  return isAPreviousCache ? self.caches.delete(cacheName) : Promise.resolve();
});
```

What do you think (I tend to delay the use of comments as much as possible). You can use `.startsWith()`. It is not ES5 but, you know, this PR is about converting into arrow functions. ;)

delapuente:
Cool, let's merge.

delapuente:
@marco-c the [build is failing](https://travis-ci.org/mozilla/oghliner/builds/103666609) but there are no differences between the HEAD of the repository and 47ce4b3 which is ok. I'm testing locally and I cannot reproduce the error with latest Node. Can you take a look?

marco-c:
@delapuente it was an intermittent issue.

delapuente:
Ok, it failed me three times in a row but it's good to hear and makes sense. ;)

PR #267, opened 2016-01-20
Assert that prepareCache returns an empty Cache

marco-c:


test/service-worker/testCacheSetup.js
```diff
@@ -29,6 +29,13 @@ describe('Setup methods for cache', function () {
       .then(function () {
         return oghliner.prepareCache();
       })
+      .then(function (cache) {
+        assert.equal(cache.constructor.name, "Cache");
```
delapuente: Try `assert.instanceOf(cache, Cache)`.

Cache constructor should be available in the ServiceWorkerGlobalScope.

marco-c: Nice! Done in 92d6e1e.

Issue #258, opened 2016-01-12
Make Service Worker tests to run on Travis


delapuente:
Make Karma run on Travis so we can pass the service worker tests in the CI environment.

marco-c:
To download the latest Firefox Nightly, we could re-use some of the code I used in my `web-push` library: https://github.com/marco-c/web-push/blob/e142930e117e433f10101f5bf5086a489945fb56/test/selenium-init.js

marco-c:
Fixed by #269.

PR #269, opened 2016-01-28
Adding Karma tests to Travis CI

delapuente:


delapuente:
This fix #258 
@marco-c, do you mind to review?
And thank you for all the previous assistance.

.travis.yml
```diff
@@ -4,21 +4,30 @@ node_js:
   - '4'
   - 'stable'
 addons:
+  firefox: 'latest-beta'
```
marco-c: Can we use `latest` (44)?

.travis.yml
```diff
@@ -4,21 +4,30 @@ node_js:
...
   apt:
     sources:
     - ubuntu-toolchain-r-test
+    - google-chrome
     packages:
     - g++-4.8
+    - google-chrome-stable
```
marco-c, line 14: Nit: We could use the `chromium-browser` package so we don't have to add the `google-chrome` PPA earlier.

marco-c:
Nice! Very simple!

delapuente:
@marco-c I don't get it to work with Chromium. Do you mind if we merge this way and see why is not working with Chromium in a follow up?

marco-c:
Sure! It was just a "nit"!

PR #279, opened 2016-03-08
Test with Chromium so we don't have to add additional apt sources and install additional packages

marco-c:


mykmelez:
Makes sense.

PR #286, opened 2018-05-22
Switches to using the "deploy" step with a "script" provider

imjaroiswebdev:


mykmelez:
Thanks for the fix, and sorry for the delay reviewing it! I'm having trouble testing it due to issues with testing Oghliner generally, so it's taking me longer than usual to review it. Are you able to test it locally, and have you had to make any other changes to do so?

imjaroiswebdev:
Hello Myk, thanks for answering me...

Well I have to tell you, I the last two nights I have been trying to make
Travis to complete the build script, since the new deploy implementation is
working, but first the problem was that sw test were failing due to karma
seems to not be able to execute the test inside the web worker scope, so I
tried disabling those test, but find that now node-gyp is not compiling the
ursa package, because a mismatch that finds when running in Node 10, but in
earlier version of Node is passing.

So, I do not want to give up with that and I have plans to solve this
weekend, first the issue of node-gyp, that is related to a breaking change
introduced in version 10 of Node and then to make a new implementation for
running the sw test scripts with the karma runner.

Because I think that without this updates the issues deploying will
continue.

Or do you have any other idea that can enligh me Myk? 😁

Greetings...

mykmelez:
> Or do you have any other idea that can enligh me Myk? 😁

I don't have any better idea! Your plan sounds like the best one. Have you been able to make progress on it?

mykmelez:
I'm going to archive this project, as it isn't being actively developed anymore. Please feel free to fork it and continue development on your fork, however.

